PR #2, opened 2017-11-24
Add indent and autopep8 infrastructure

tiran:
* indent is a tool to auto-format C code. The ``.indent.pro`` file is based on CPython's ``Misc/indent.pro`` file. It should get us very close to PEP 7 compatible C code.
* autopep8 is a Python library to auto-format Python code. It can be used to address single pycodestyle violation, e.g. ``E121 - Fix indentation to be a multiple of four.``. See https://pypi.python.org/pypi/autopep8#features

### DIFF
--- a/.indent.pro
+++ b/.indent.pro
@@ -1,0 +1,24 @@
+--blank-lines-after-declarations
+--blank-lines-after-procedures
+--braces-after-func-def-line
+--braces-on-if-line
+--braces-on-struct-decl-line
+--break-after-boolean-operator
+--comment-indentation25
+--comment-line-length79
+--continue-at-parentheses
+--dont-cuddle-do-while
+--dont-cuddle-else
+--indent-level4
+--line-length79
+--no-space-after-casts
+--no-space-after-function-call-names
+--no-space-after-parentheses
+--no-tabs
+--procnames-start-lines
+--space-after-for
+--space-after-if
+--space-after-while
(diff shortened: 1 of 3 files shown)


encukou:
Thanks.
I don't think installing tools from the `Makefile` is a good idea, though. Would you be OK limiting that to just running the tool?
The installation could be documented separately (but not necessarily; `autopep8: command not found` is quite an actionable error message).

tiran:
Ah, Fedora has ``python3-autopep8`` package. Good, that makes our job a but easier.

encukou:
Let's worry about this after 3.0.

jdufresne:
I think it would be good to land this change before releasing 3.0.0. A new major release is the best time to do large project wide changes. IIUC, this will modify the code but should not have any side effects. Having more idiomatic Python code as well as avoiding inadvertent whitespace changes will help make new contributions easier.

encukou:
3.0.0 is mainly about merging contributors from pyldap, which, being a fork that was meant to be merged back to python-ldap, avoided reformatting the code.
Helping make new contributions easier is the next step, but please wait for after the release. (It should be here soon; there's one bug to fix!)

jdufresne:
Makes sense. Thanks for the response.

encukou:
Hi,
I've rebased, added an `autoformat` target to hide the implementation details, and mentioned this in the Contributing section of the docs. Does this look good?

PR #5, opened 2017-11-24
Format the README

encukou:
* Add a blank line before the list, so ReST treats it as a list
* Remove e-mail addresses
* Just make it UTF-8 and wait if someone complains – I munged Miro's name in an effort to keep it ASCII, but forgot an `ö` and others...
* Workaround for an initial

### DIFF
--- a/README
+++ b/README
@@ -102,25 +102,25 @@ These very kind people have supplied patches or suggested changes:
 
 These people contributed to Python 3 porting (at https://github.com/pyldap/):
 
-* A. Karl Kornel <akkornel at stanford.edu>
-* Alex Willmer <alex at moreati.org.uk>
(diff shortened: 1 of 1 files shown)


encukou:
The name `A. Karl Korner` is, unfortunately, not friendly with ReStructured Text  syntax :(

The initial "A." makes think it's a beginning of an ordered list, with headings A., B., C., etc. The result looks bad and is semantically wrong.

The standard way to escape is using a backslash, "A\.". This looks bad when viewing the source.

Another way is to use a zero-width space. This is invisible when rendered and in some editors, but more paranoid text-oriented tools render it as "<U+200B>A." (less) or "<200b>A." (vim).
But, it's outside the name, which makes it a better solution than "A\."

@akkornel, if you would like to use your full first name instead, let me know.

PR #7, opened 2017-11-24
Add GCOV / LCOV C coverage support

tiran:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,13 @@
 language: python
 
+python:
+- '2.7'
+- '3.3'
+- '3.4'
+- '3.5'
+- '3.6'
+# Note: when updating Python versions, also change setup.py and tox.ini
+
 sudo: false
 
 cache: pip
@@ -10,28 +18,8 @@ addons:
       - ldap-utils
       - slapd
 
-matrix:
-  include:
-    - python: 2.7
-      env:
-        - TOXENV=py27
-        - WITH_GCOV=1
-    - python: 3.3
-      env:
-        - TOXENV=py33
-        - WITH_GCOV=1
-    - python: 3.4
-      env:
-        - TOXENV=py34
-        - WITH_GCOV=1
-    - python: 3.5
-      env:
-        - TOXENV=py35
-        - WITH_GCOV=1
-    - python: 3.6
-      env:
-        - TOXENV=py36
-        - WITH_GCOV=1
+env:
+  - WITH_GCOV=1
 
 install:
   - pip install "pip>=7.1.0"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -9,3 +14,26 @@ clean:
...
+$(LCOV_REPORT): $(LCOV_INFO)
+	genhtml --output-directory $(LCOV_REPORT) \
+		$(LCOV_REPORT_OPTIONS) $(LCOV_INFO)
+
+lcov-report: $(LCOV_REPORT)
+
+lcov-open: $(LCOV_REPORT)
+	xdg-open $(LCOV_REPORT)/index.html
+
+lcov: lcov-clean lcov-coverage lcov-report
```
encukou: I don't think make guarantees the order in which it builds dependencies (especially with parallel builds – I have `make` aliased to `make -j4`).

IMO it should be something like:

```
lcov: lcov-clean
    $(MAKE) lcov-coverage
    $(MAKE) lcov-report
```

.travis.yml
```diff
@@ -18,12 +10,37 @@ addons:
       - ldap-utils
       - slapd
 
+matrix:
+  include:
+    - python: 2.7
+      env:
+        - TOXENV=py27
+        - WITH_GCOV=1
```
encukou: Is `WITH_GCOV` necessary here? It's specified in `tox.ini`.

Makefile
```diff
@@ -9,3 +14,26 @@ clean:
 	rm -f .coverage .coverage.*
 	find . -name '*.py[co]' -or -name '*.so*' -or -name '*.dylib' -delete
 	find . -depth -name __pycache__ -exec rm -rf {} \;
+
+# LCOV report
```
encukou: This will send anyone unfamiliar `LCOV` to Google. Can you note what this is? e.g. `# LCOV report (measuring test coverage for C code)`

.travis.yml
encukou: Oh, passenv only passes through existing variables, right? In that case I guess it's fine.

tiran: tox.ini has ``passenv = WITH_GCOV``. It does not set ``WITH_GCOV`` but let tox pass the env vars through. By default tox unsets most env vars.

Makefile
tiran: the makefile has ``.NOTPARALLEL:`` to disable all parallel builds. But I can change the code.

encukou:
I found a way to simplify the config. Does this look good to you?

(Sorry for the mess I made when pushing it to this PR)

tiran:
ACK

We can move to matrix configuration later in order to run CI for linters and documentation.

PR #8, opened 2017-11-24
Check for LDAP schema and slapd binaries

tiran:
Integration tests depend on openldap client and server binaries. Fail
with a human readable message in case a tool or schema is missing.

### DIFF
--- a/Lib/slapdtest.py
+++ b/Lib/slapdtest.py
@@ -115,7 +115,7 @@ class SlapdObject(object):
     elif os.path.isdir("/etc/ldap/schema"):
         SCHEMADIR = "/etc/ldap/schema"
     else:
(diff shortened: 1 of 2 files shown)


encukou:
What's your use case?
I'd want the CI tests to go red if something is misconfigured this badly.

tiran:
I can change the PR to print a useful error message instead of skipping the tests. At the moment the test suite fails with cryptic error messages when schema or client binaries are missing.

encukou:
Thanks!
In the future we might want to make a distinction between skipping tests with unmet dependencies, and skipping based on the test matrix (e.g. bytes_mode on Python 3). But CI should always fail if misconfigured.

PR #11, opened 2017-11-24
Build documentation without the compiled extension

encukou:
This rabbit hole turned out deeper than expected, but I managed to mock out `_ldap` so that documentation can be built without a C compiler.
All the magic is documented.

C code had the only list of `LDAP_*` constants python-ldap used. I turned that into a Python file, and generated a `.h` from it.


encukou:
Autodoc is working as it should, see e.g. [ldap-controls](http://python-ldap.readthedocs.io/en/read-the-docs/ldap-controls.html)

tiran:
I'll review this PR later today. It's rather big.

Modules/constants.c
```diff
@@ -6,373 +6,180 @@
...
+int
+LDAPinit_constants( PyObject* m )
+{
+    PyObject *exc;
+    int result;
 
-  obj = PyUnicode_FromString(LDAP_CONTROL_ASSERT);
-  PyDict_SetItemString( d, "CONTROL_ASSERT", obj );
-  Py_DECREF(obj);
+#define check_result() {  \
```
tiran: I don't like this kind of macros. It's confusing whenever a macro picks up values from the outer scope. Please either pass result in the function ``check(result);``or rewrite the few calls to ``if (PyModule_AddIntConstant(m, "OPT_ON", 1)) return -1;``

IIRC some compilers do not like a semicolon after a block ``{...};``. This could be a problem with MSVC.

Lib/ldap/constants.py
```diff
@@ -0,0 +1,402 @@
...
+    Str("CONTROL_SYNC_STATE"),
+    Str("CONTROL_SYNC_DONE"),
+    Str("SYNC_INFO"),
+    Str("CONTROL_PASSWORDPOLICYREQUEST"),
+    Str("CONTROL_PASSWORDPOLICYRESPONSE"),
+    Str("CONTROL_RELAX"),
+)
+
+
+def write_header():
```
tiran: It does a bit more than just writing the header :)

Modules/constants.c
```diff
@@ -6,373 +6,180 @@
...
+    PyErr_SetNone(errobjects[errnum+LDAP_ERROR_OFFSET]);
+  else
+    PyErr_SetObject(LDAPexception_class,
+    Py_BuildValue("{s:i}", "errnum", errnum));
+  return NULL;
+}
+
+/* Convert an LDAP error into an informative python exception */
+PyObject*
+LDAPerror( LDAP *l, char *msg )
```
tiran: Is this a straight copy from the other file?

Lib/ldap/constants.py
```diff
@@ -0,0 +1,402 @@
+"""Definitions for constants exported by OpenLDAP
```
tiran: Can you add a shebang, please?

encukou: Should I? As a Linux packager, I have a bias against executable files in `/usr/lib`.
The module is intended to be run as `python Lib/ldap/constants.py > Modules/constants_generated.h`, and that doesn't need a shebang.

Lib/ldap/constants.py
encukou: I don't understand the comment. It dumps to stdout in C header format, nothing more :)
I've renamed it to `print_header` to point out the output is not configurable.

Modules/constants.c
encukou: Yes, except whitespace and a comment.

    git diff HEAD^{/Refactor}~:Modules/errors.c HEAD^{/Refactor}:Modules/constants.c

Modules/constants.c
encukou: OK, I'll write out the error checking. (but I prefer an explicit `!= 0` if what I'm checking isn't a straightforward Boolean.)

Wrapped the macros in `do {...} while (0)`.

Lib/ldap/constants.py
tiran: Ah, you refer to C header. I though that you were referring to file header `` Generated with: ...``. Forget what I said. :)

Modules/constants.c
tiran: +1

(cool git magic)

tiran:
Awesome, LGTM! 💯

PR #12, opened 2017-11-24
Fix formatting of README

spaceone:
The authors weren't shown as list with bullet points but with unformatted '*' .
Btw, I think you are doing a great job.


encukou:
Thanks for helping!

There's a PR open for this already though; just waiting to hear if it's OK to re-format a name: https://github.com/python-ldap/python-ldap/pull/5

PR #15, opened 2017-11-25
Run tests with warnings and bytes warnings

tiran:
All warnings except expected warnings are turned into exceptions.

Fixed a resource warning in schema test case.


encukou:
I'm worried that I don't see a good process to remove warning filters once they become obsolete. I'm afraid I don't have a solution for that, though.
Somewhat related, I'd like to see some comment for each filter. It's not at all obvious why python-ldap would use the `imp` module.
As for warnings that python-ldap itself raises, is that really something to ignore? Or should we rather change the tests – perhaps even test that the warning is raised?

I'm not saying all the work needs to be done now, but if not, let's document the direction this should take.

tiran:
I'd really want to enable BytesWarning with Python 3 to fetch bugs like ``str(b'data')``. We can safely ignore warnings in Python 2.7 runs for now, and enable them later when we have a better way to deal with warnings then:

```
    "-Wignore:Under Python 2, python-ldap uses bytes:BytesWarning" \
    "-Wignore:Received non-bytes value:BytesWarning" \
```

I was able to get rid of ``ImportWarning`` by renaming a directory. The deprecation warning occurs outside of python-ldap. distutils imports the ``imp`` module internally. I also added a couple of comments to ``tox.ini``.

Issue #18, opened 2017-11-26
Resource leak in error case of LDAPControls_from_object()


tiran:
The error case in ``LDAPControls_from_object(clientctrls, &client_ldcs)`` does not clean up ``server_ldcs`` in multiple places. It should call ``LDAPControl_List_DEL( server_ldcs );``. Some function also do not call ``free_attrs( &attrs, attrs_seq);`` in error cases, too.

* https://github.com/python-ldap/python-ldap/blob/python-ldap-2.5.2/Modules/LDAPObject.c#L627
* https://github.com/python-ldap/python-ldap/blob/python-ldap-2.5.2/Modules/LDAPObject.c#L691
* https://github.com/python-ldap/python-ldap/blob/python-ldap-2.5.2/Modules/LDAPObject.c#L751
* more...

PR #21, opened 2017-11-26
Fix resource leaks in error cases

tiran:
On several occasions LDAPObject methods did not clean up LDAPControls
and other resources when LDAPControls_from_object() fails. This bug
would lead to memory leaks in case a server control or client control
could not be handled correctly.

Closes: #18

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -361,8 +361,10 @@ l_ldap_unbind_ext( LDAPObject* self, PyObject* args )
     }
 
     if (!PyNone_Check(clientctrls)) {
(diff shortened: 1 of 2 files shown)


encukou:
Squashed and merged as a72aaace0c4bd1085d23bb740fbc1183c7a16aba (with tiny a whitespace fix)

tiran:
@encukou This fix has not been merged into master. a72aaac is a different commit.

encukou:
Oh dear. My apologies!

I see you re-opened as https://github.com/python-ldap/python-ldap/pull/36; I merged that.

PR #28, opened 2017-11-27
Omit SysLogHandler when /dev/log does not exist

tiran:
slaptests use SysLogHandler() to write logs to syslog. Don't create a
SysLogHandler when /dev/log does not exist. This fixes a bug when
building and testing python-ldap in containers or restricted build
environments.

### DIFF
--- a/Lib/slapdtest.py
+++ b/Lib/slapdtest.py
@@ -59,7 +59,7 @@ def combined_logger(
             pass
     # for writing to syslog
     new_logger = logging.getLogger(log_name)
-    if sys_log_format:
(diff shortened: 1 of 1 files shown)


encukou:
This section looks fishy to me. The [SysLogHandler documentation](https://docs.python.org/3/library/logging.handlers.html#sysloghandler) says the address is system-dependent (e.g. `/var/run/syslog` on macOS).

I'm not opposed to merging a quick fix for a hack, but please add a `XXX` comment or file an issue in that case.

tiran:
We could just remove the ``SysLogHandler`` completely. It's only used in verbose test mode. I don't think that tests should spam the system log.

Issue #34, opened 2017-11-27
Use explicit registration for ldap.schema.subentry.SCHEMA_CLASS_MAPPING


encukou:
The current code does `import *` and then picks classes with `schema_attribute` set.
Be more explicit here – e.g. use a decorator.

See https://github.com/python-ldap/python-ldap/pull/25

spaceone:
Another possibility would be a metaclass which registers them.

PR #35, opened 2017-11-27
Lib: Avoid eval() for getting module-level variables

encukou:
Alternate version of #25 that avoids `eval()` entirely.

Issue for a more proper fix filed here: https://github.com/python-ldap/python-ldap/issues/34

Fixes: https://github.com/python-ldap/python-ldap/pull/25

### DIFF
--- a/Lib/ldap/schema/subentry.py
+++ b/Lib/ldap/schema/subentry.py
@@ -11,8 +11,7 @@ from ldap.schema.models import *
 SCHEMA_CLASS_MAPPING = ldap.cidict.cidict()
 SCHEMA_ATTR_MAPPING = {}
 
-for _name in dir():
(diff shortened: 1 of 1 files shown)


tiran:
LGTM

(The ``list()`` call is properly not necessary. It doesn't hurt either. Keep it)

encukou:
Actually, it is necessary: the `vars()` dict changes when `o` is added to the namespace :)
(Not that I would figure that out without tests...)

AlanCoding:
Confirmed that this fixes the traceback in our environment, which was described in https://mail.python.org/pipermail/python-ldap/2016q3/003789.html

Thank you. We will be able to update our dependencies when this goes out in a release.

tiran:
Oh fun :/

PR #48, opened 2017-11-28
Doc: Link to documentation for older versions

encukou:
`python-ldap.org` includes a link to pre-built documentation (although it was not updated recently). This will be lost with the switch to Read the Docs – it's not set up to serve zipfiles, and older branches don't build there.
I've built docs for latest 2.4.x and 2.5.x releases and uploaded them to GitHub. This PR makes the documentation point there.


tiran:
The zipfile contains only the root directory but no files or subdirectories.

https://github.com/python-ldap/python-ldap/releases/download/python-ldap-2.5.2/python-ldap-docs-html-2.5.2.zip

encukou:
Fixed now. Thanks for catching that!

PR #49, opened 2017-11-28
Fix several compiler warnings and make warnings fatal errors

tiran:
* Correct return LDAPerror_TypeError(). LDAPerror_TypeError() returns ``(PyObject *)NULL``. Some functions don't return a ``PyObject*`` so technically it's wrong to return ``(PyObject *)NULL``. Return NULL instead.
* Fix implicit declaration of error functions, ldapcontrol.c was missing include of constants.h that defines LDAPerr() and LDAPerror().
* Make init_ldap_module definition a strict prototype
* Travis: Turn compiler warnings into fatal errors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,14 @@ matrix:
     - python: 3.6
       env: TOXENV=doc
 
+env:
+    global:
+        # -Wno-int-in-bool-context: don't complain about PyMem_MALLOC()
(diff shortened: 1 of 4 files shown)


encukou:
Code changes LGTM, but see https://github.com/python-ldap/python-ldap/pull/53 for the CFLAGS setting.

PR #52, opened 2017-11-29
Fix schema urlfetch and add tests

tiran:


tiran:
@encukou The second commit addresses the issue with urlfetch on Python 2.7. I don't know if it's the right way to resolve the issue. It works for me, though. Please have a look.

encukou:
I believe that's the right fix. Thanks!

PR #53, opened 2017-11-29
Fix declaration after statement for C90

tiran:
In strict ISO C90 mode and on some platforms (e.g. older MSVC), it is
not allowed to declare new variables after a statement. I also find it
good practice to declare first.

Fixes error: ISO C90 forbids mixed declarations and code
[-Werror=declaration-after-statement]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,8 @@ env:
     global:
         # -Wno-int-in-bool-context: don't complain about PyMem_MALLOC()
         # -Werror: turn all warnings into fatal errors
-        - CFLAGS="-Wno-int-in-bool-context -Werror"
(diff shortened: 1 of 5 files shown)


encukou:
This only sets CFLAGS for Travis.
I assume that's because they are `gcc`-specific, and setting them correctly in `setup.py` is unreasonably hard. Is that right?

tiran:
I don't want to get in the business of detecting compiler flags in setup.py. It is possible but lots of extra work. Or do you want to include autoconf?

encukou:
That's reasonable.

tiran:
Oh fun, now Travis is failing with more warnings. Please merge #49 first.

Issue #60, opened 2017-11-29
CONNECT_ERROR in 3rd TLS handshake on Fedora (NSS backend)


tiran:
PR #42 introduced a test for SASL EXTERNAL auth with TLS client certs. The test case works fine on Travis (Ubuntu). It's also passing on Fedora 27 when executed alone. But when executed with the rest of the test case, slapd refuses STARTTLS with CONNECT_ERROR. PR #59 disables the test temporarily.

I need to investigate why the test is failing. It might be related to the fact that Fedora uses NSS instead of OpenSSL.

tiran:
I was able to track this down to a problem with NSS layer in openldap. It looks like ``SECMOD_RestartModules()`` call fails. I also wonder why ``tlsm_deferred_ctx_init`` is called again. It's protected by ``PR_CallOnceWithArg()`` so it should only be initialized once.

```
Breakpoint 2, alloc_handle (ctx_arg=0x55f297b0e480, is_server=0) at tls2.c:280
280     {
(gdb) n
284             if ( ctx_arg ) {
(gdb) p ctx
$6 = <optimized out>
(gdb) n
292             ssl = tls_imp->ti_session_new( ctx, is_server );
(gdb) p tls_imp
$7 = <optimized out>
(gdb) p tls_imp->ti_session_new
value has been optimized out
(gdb) s
tlsm_session_new (ctx=0x55f297b0e480, is_server=0) at tls_m.c:2769
warning: Source file is more recent than executable.
2769    {
(gdb) n
2776            c->tc_is_server = is_server;
(gdb) n
2777            LDAP_MUTEX_LOCK( &tlsm_init_mutex );
(gdb) n
2778            status = PR_CallOnceWithArg( &c->tc_callonce, tlsm_deferred_ctx_init, c );
(gdb) n
2779            LDAP_MUTEX_UNLOCK( &tlsm_init_mutex );
(gdb) n
2778            status = PR_CallOnceWithArg( &c->tc_callonce, tlsm_deferred_ctx_init, c );
(gdb) n
2779            LDAP_MUTEX_UNLOCK( &tlsm_init_mutex );
(gdb) n
2780            if ( PR_SUCCESS != status ) {
(gdb) p status
$8 = PR_FAILURE
```

Next session

```
Breakpoint 3, tlsm_deferred_ctx_init (arg=0x55f297b0e480) at tls_m.c:2329
2329    {
(gdb) n
[...]

PR #62, opened 2017-11-30
[WIP] Add reproducer for start_tls issue with NSS

tiran:
See https://github.com/python-ldap/python-ldap/issues/60


encukou:
Do we need this in the repository?
This seems better suited for a pastebin, e.g. (Gist)[https://gist.github.com/]

tiran:
I'll convert it to a test case once the issue has been resolved.

encukou:
Alright; if you don't need it in master then let's keep it WIP

tiran:
Test case is in https://github.com/python-ldap/python-ldap/pull/134

PR #63, opened 2017-11-30
Doc: Add dedicated page for bytes_mode

encukou:
This moves the bytes_comment from the main page of the docs.


encukou:
It seems I have a different dictionary locally...

tiran:
The dictionary on Travis might be a bit dated. It's an older version of Ubuntu.

Issue #55, opened 2017-11-29
[doc] Explain TLS/SSL gotchas


tiran:
TLS/SSL and X.509 are tricky. OpenLDAP adds some additional gotchas to the stack. We should document them so users don't run into issues like https://github.com/pyldap/pyldap/issues/53

* ``start_tls_s()`` requires protocol version 3. It does not work with protocol version 2.
* Several, perhaps all ``set_option(OPT_X_TLS_*, ...)`` calls require a final ``set_option(ldap.OPT_X_TLS_NEWCTX, 0)`` call to submit all previous ``set_option()`` calls. Without ``OPT_X_TLS_NEWCTX``, settings are effectively ignored. 

```
        l.set_option(ldap.OPT_PROTOCOL_VERSION, ldap.VERSION3)
        l.set_option(ldap.OPT_X_TLS_CACERTFILE, 'path/to/ca.pem')
        l.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap. OPT_X_TLS_NEVER)
        l.set_option(ldap.OPT_X_TLS_NEWCTX, 0)
```

tiran:
See https://bugzilla.redhat.com/show_bug.cgi?id=1048475

encukou:
Do we want to actually track this in code, issue a warning for one release, then make it a hard error?

tiran:
It's a bit tricky. The cert settings can be changed globally ``ldap_set_option(NULL, opt, val)`` and locally ``ldap_set_option(conn, opt, val)``. I'm not sure how to fix cases like:

```
conn = ldap.initialize()
# now change the global setting but expect the local connection to pick up the change
ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, True)
conn.start_tls_s()
```

encukou:
Documenting it is fine, or at least a good first step. We don't need to hand-hold the users (though it would be nice if we did).

encukou:
Anyway, I don't think 3.0.0 beta needs to wait for this, as it's not a regression from either python-ldap or pyldap.

graingert:
https://github.com/tribe29/checkmk/commit/caa7fe1c16393322bcc12b13333d48ff59eedd09

sometimes settings OPT_X_TLS_NEWCTX throws a ValueError, but it still connects successfully

This is a problem for reconnecting ldap object because it doesn't catch the value error when settings OPT_X_TLS_NEWCTX 

@LarsMichelsen
[...]

PR #64, opened 2017-11-30
[WIP] Check and warn if OPT_X_TLS_NEWCTX is required

tiran:
See https://github.com/python-ldap/python-ldap/issues/55


encukou:
What's the progress here? Is it a dead end?

tiran:
The documentation improvements are now merged. Let's not do the extra checks.

PR #69, opened 2017-12-01
Prepare 3.0.0 Beta

encukou:


CHANGES
```diff
@@ -60,21 +73,31 @@ Lib/
 Lib/slapdtest.py
 * Automatically try some common locations for SCHEMADIR
 * Ensure server is stopped when the process exits
+* Check for LDAP schema and slapd binaries
```
tiran: slapdtest also includes certificates (CA, server, client) and configuration for TLS client auth.

Apropos certificates, the test certs are **not** installed with ``slapdtest``. Is slapdtest shipped in standard distributions?

encukou: It is in the sdist, yes.

encukou: But actually, it's python-ldap's test suite that includes the certificates, not not `slapdtest.py`.
This should be fine.

encukou: Fixing in https://github.com/python-ldap/python-ldap/pull/73

tiran: thanks, I opened #76 to track the issue.

PR #71, opened 2017-12-01
Doc: Add a separator between sentences

hroncok:
This rendered as follows:

> OpenLDAP client libs version 2.4.11 or later It is not possible and not supported to build with prior versions.


encukou:
Thank you!

PR #73, opened 2017-12-01
Add Tests/cert to MANIFEST.in

encukou:


tiran:
slapdtest assume that current working directory contains ``Tests/certs/ca.pem`` etc. It will no longer work outside of the test tree.

Your fix is good enough for beta. For final release I have to come up with a better solution. I wasn't aware that we actually ship the test helper as part of the standard distribution and install it in site-packages.

encukou:
Oh. Yeah, I missed that :(

This PR doesn't really solve any issue, then.

tiran:
Without this PR, it is not even possible to run tox from a sdist.

PR #77, opened 2017-12-02
Ship documentation in source distribution

tiran:
Source distributions (sdist) of python-ldap now contain documentation.
This allows users to build and read docs locally. Also fixes tox test
run in sdist.


tiran:
Ah, good one :)

Issue #76, opened 2017-12-01
Properly handle certs in slapdtest


tiran:
slapdtest now depends on certificates and keys for STARTTLS and SASL tests. The files are not available unless slapdtest is started from a working directory in the root of a python-ldap distribution. We either need to make certs optional or ship the certs as package data with ldap Python package.

See #73

PR #78, opened 2017-12-02
 Convert slapdtest to package with data

tiran:
The slapdtest module has been converted to a package to ship the test certs as package data.

Prolong life time of certs.
    
Closes: https://github.com/python-ldap/python-ldap/issues/76


encukou:
This is almost what I did offline over the weekend. Let me merge in my changes & rebase.

encukou:
The biggest change is moving code into a submodule – logic in `__init__.py` generally leads to horror stories over time.

tiran:
Any suggestions? Do you want to move ``SlapdObject`` and ``SlapdTestCase`` to a module and import them into the package? Different module for the test helper functions?

encukou:
Put all the code in one private submodule, which can be split later. And import public API into the package. (For "public API" I used what we currently need outside slapdtest; it can be expanded later.)

See this PR. (Best viewed in a local clone; GitHub's timestamp-based order of the commits is confusing.)

Lib/slapdtest/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+"""
+slapdtest - module for spawning test instances of OpenLDAP's slapd server
+
+See https://www.python-ldap.org/ for details.
+"""
+
+__version__ = '3.0.0b1'
+
+from slapdtest._slapdtest import SlapdObject, SlapdTestCase, SysLogHandler
+from slapdtest._slapdtest import skip_unless_ci, requires_tls
```
tiran: +1

The test helpers could be moved into a separate module.

PR #83, opened 2017-12-04
Add valgrind target to check for memory leaks

tiran:


encukou:
I'll look at this one after 3.0 beta.

encukou:
The `python3` and the path suppression are Fedora-specific. That's as good a default as any, but it needs to be readily configurable.
I tried fixing this, does that look OK?

tiran:
Use the makefile force, Luke :)

```
$(PYTHON_SUPP):
    @echo "Need support file"
    exit 1

valgrind: $(PYTHON_SUPP)
```

encukou:
Ooh, shiny! Glad to be your apprentice :)

Issue #79, opened 2017-12-04
Memory leaks in  ldap_create_assertion_control_value


tiran:
The function `` ldap_create_assertion_control_value`` leaks a BER data structure and a ``LDAP*`` structure.

```
==29355== 14 bytes in 1 blocks are definitely lost in loss record 15 of 5,810
==29355==    at 0x4C2FB6B: malloc (vg_replace_malloc.c:299)
==29355==    by 0x14CBD9E4: ber_memalloc_x (memory.c:228)
==29355==    by 0x14CBC89D: ber_flatten2 (io.c:407)
==29355==    by 0x14AA026F: ldap_create_assertion_control_value (assertion.c:58)
==29355==    by 0x1485C427: encode_assertion_control (ldapcontrol.c:351)
==29355==    by 0x4FC21AE: _PyCFunction_FastCallDict (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4FC9279: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4FF2EC9: _PyEval_EvalFrameDefault (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4F8BBE9: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4FC933D: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4FF2EC9: _PyEval_EvalFrameDefault (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4F8BBE9: ??? (in /usr/lib64/libpython3.6m.so.1.0)
```

```
==29355== 33,691 (40 direct, 33,651 indirect) bytes in 1 blocks are definitely lost in loss record 5,807 of 5,810
==29355==    at 0x4C31A1E: calloc (vg_replace_malloc.c:711)
==29355==    by 0x14CBDA54: ber_memcalloc_x (memory.c:283)
==29355==    by 0x14A77010: ldap_create (open.c:115)
==29355==    by 0x1485C40D: encode_assertion_control (ldapcontrol.c:347)
==29355==    by 0x4FC21AE: _PyCFunction_FastCallDict (in /usr/lib64/libpython3.6m.so.1.0)
==29355==    by 0x4FC9279: ??? (in /usr/lib64/libpython3.6m.so.1.0)
[...]

Issue #86, opened 2017-12-04
encode_assertion_control creates temporary LDAP* struct


tiran:
The function ``encode_assertion_control`` creates a temporary LDAP* struct. This seems pretty wasteful and causes a memory leak #79. We should find a simpler way to encode assertion controls.

https://github.com/python-ldap/python-ldap/blob/c84b4faf429431cb4d4ef42a156b62d1105f1758/Modules/ldapcontrol.c#L333-L360

PR #88, opened 2017-12-04
Fix two memory leaks in encode_assertion_control

tiran:
encode_assertion_control was leaking a LDAP* struct and BER data values
on every call. Both data structures are now properly freed and the GIL
is released around ldap_create() and ldap_unbind_s().

Closes: https://github.com/python-ldap/python-ldap/issues/79

### DIFF
--- a/Modules/ldapcontrol.c
+++ b/Modules/ldapcontrol.c
@@ -343,16 +343,33 @@ encode_assertion_control(PyObject *self, PyObject *args)
         goto endlbl;
     }
 
+    /* XXX: ldap_create() is a nasty and slow hack. It's creating a full blown
+     * LDAP object just to encode assertion controls.
+     */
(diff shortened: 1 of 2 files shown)


Modules/ldapcontrol.c
```diff
@@ -343,16 +343,33 @@ encode_assertion_control(PyObject *self, PyObject *args)
         goto endlbl;
     }
 
+    /* XXX: ldap_create() is a nasty and slow hack. It's creating a full blown
+     * LDAP object just to encode assertion controls.
+     */
```
encukou: I took the liberty to mark the comment with `XXX`.

tiran: The issue is tracked in https://github.com/python-ldap/python-ldap/issues/86

PR #90, opened 2017-12-04
Fix set_option() of timeout to handle -1 as infinity

tiran:
Also adds a bunch of tests for options.


encukou:
I split the first commit into two.
Also, I changed `infinitive` to `infinity` in some comments. (“Infinitive” is [a completely unrelated term from linguistics](https://en.wikipedia.org/wiki/Infinitive); the adjective form would be “infinite”.)

Issue #100, opened 2017-12-05
conn.set_option(-1, '') does not fail sometimes


tiran:
I started to create tests for ``SimpleLDAPObject.set_option`` and noticed something strange. Sometimes a ``set_option`` call does not fail although it should fail, e.g. ``conn.set_option(-1, '')`` should fail with ``ValueError: unknown option -1``.

While I was investigating the issue with pdb, I ran into failing pdb session. pdb's interactive session just failed with ``RuntimeWarning: tp_compare didn't return -1 or -2 for exception``:

```
........> /home/heimes/dev/redhat/python-ldap/Tests/t_ldap_options.py(141)test_invalid()
-> self.conn._l.set_option(-1, '')
(Pdb) n
--Return--
/usr/lib64/python2.7/bdb.py:351: RuntimeWarning: tp_compare didn't return -1 or -2 for exception
  i = max(0, len(stack) - 1)
E
======================================================================
ERROR: test_invalid (__main__.TestLDAPObjectOptions)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "Tests/t_ldap_options.py", line 141, in test_invalid
    self.conn._l.set_option(-1, '')
  File "/usr/lib64/python2.7/bdb.py", line 53, in trace_dispatch
    return self.dispatch_return(frame, arg)
  File "/usr/lib64/python2.7/bdb.py", line 88, in dispatch_return
    self.user_return(frame, arg)
  File "/usr/lib64/python2.7/pdb.py", line 190, in user_return
    self.interaction(frame, None)
  File "/usr/lib64/python2.7/pdb.py", line 208, in interaction
    self.setup(frame, traceback)
  File "/usr/lib64/python2.7/pdb.py", line 119, in setup
    self.stack, self.curindex = self.get_stack(f, t)
  File "/usr/lib64/python2.7/bdb.py", line 351, in get_stack
    i = max(0, len(stack) - 1)
ValueError: unknown option -1

----------------------------------------------------------------------
Ran 9 tests in 8.825s
```

Python 3 doesn't like  ``conn.set_option(-1, '')`` either:

```
ValueError: unknown option -1

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "Tests/t_ldap_options.py", line 140, in test_invalid
    self.conn._l.set_option(-1, '')
SystemError: <built-in method set_option of LDAP object at 0x7f01fdb16bc0> returned a result with an error set
```

Issue #95, opened 2017-12-04
timeval options do not have symmetric set/get 


encukou:
Issue description:

Infinite timeout is set using a negative number, but turns into `None` when retreived.
Setting timeout to `None` leads to an internal error.

```pycon
>>> import ldap
>>> conn = ldap.initialize('https://example.com')
>>> conn.set_option(ldap.OPT_NETWORK_TIMEOUT, -1)
>>> print(conn.get_option(ldap.OPT_NETWORK_TIMEOUT))
None
>>> conn.set_option(ldap.OPT_NETWORK_TIMEOUT, None)
Traceback (most recent call last):
  ...
SystemError: <built-in method set_option of LDAP object at 0x7f943715dd78> returned a result with an error set
```

Operating system: Fedora
Python version: 3.6.3
python-ldap version: 3.0.0b1
Imported from https://sourceforge.net/p/python-ldap/bugs/48/

tiran:
Since https://github.com/python-ldap/python-ldap/pull/90/commits/dcb5c0052b5f83508c1c406d8ad0b23e4f5a3e6a it is possible to set timeout to -1 to indicate infinity. In C I can also map None to -1. It's just a couple of lines.

tiran:
``SystemError: <built-in method set_option of LDAP object at 0x7f943715dd78> returned a result with an error set`` is caused by issue #100 and fixed by #101. I'll implement ``None -> -1`` in another PR.

PR #101, opened 2017-12-05
Fix error reporting of LDAPObject.set_option()

tiran:
The method LDAPObject.set_option() did not signal an exception in case
the set option failed. The bug was caused by checking for the wrong
error value. The internal C function LDAP_set_option() returns ``1`` for
sucess and ``0`` for error, but LDAPObject.set_option() was checking for
``-1``.

Add some tests for LDAPObject.set_option() and LDAPObject.get_option().

Closes: https://github.com/python-ldap/python-ldap/issues/100

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -1273,7 +1273,7 @@ l_ldap_set_option(PyObject* self, PyObject *args)
 
     if (!PyArg_ParseTuple(args, "iO:set_option", &option, &value))
         return NULL;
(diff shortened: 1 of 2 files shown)


encukou:
Nice, thanks!

Just yesterday I was teaching beginners about the perils of subclassing, so I couldn't resist fixing the inheritance a bit.

tiran:
@encukou +1

I'll use the enhancements to implement test for #95 later.

Issue #99, opened 2017-12-05
Use custom warning class for BytesWarning?


tiran:
In Python 2, python-ldap emits BytesWarnings in two places unless bytes_mode is explicitly configured.

https://github.com/python-ldap/python-ldap/blob/16e8529a5cb50417ed41a89311c48b891c9e78d8/Lib/ldap/ldapobject.py#L80-L92

https://github.com/python-ldap/python-ldap/blob/16e8529a5cb50417ed41a89311c48b891c9e78d8/Lib/ldap/ldapobject.py#L122-L127

The warnings can interfere with software that has BytesWarning enabled or even turned into errors, e.g. ``python -bb`` or ``-Werror``. There is currently no stable way to suppress only BytesWarning from python-ldap. The best I came up with are two separate warning filters that depend on the content of the warning strings from python-ldap:

```
warnings.filterwarnings(
    action="ignore",
    message="Under Python 2, python-ldap uses bytes",
    category=BytesWarning
)
warnings.filterwarnings(
    action="ignore",
    message="Received non-bytes value",
    category=BytesWarning
)
```

I suggest to introduce a warnings subclass ``ldap.LDAPBytesWarning``, so users can explicitly suppress only warnings from python-ldap with 

```
import ldap
import warnings

warnings.simplefilter('ignore', ldap.LDAPBytesWarning)
```

encukou:
Sounds good, and if it's a subclass it shouldn't break anything.

PR #103, opened 2017-12-05
Use custom ldap.LDAPBytesWarning class

tiran:
Under Python 2, python-ldap now uses custom ldap.LDAPBytesWarning
instead of BytesWarning to emit warnings in default bytes mode.

Closes: https://github.com/python-ldap/python-ldap/issues/99

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -564,6 +564,18 @@ The above exceptions are raised when a result code from an underlying API
 call does not indicate success.
 
 
+Warnings
+========
+
+.. py:class:: LDAPBytesWarning
+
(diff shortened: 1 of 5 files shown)


Doc/reference/ldap.rst
```diff
@@ -564,6 +564,16 @@ The above exceptions are raised when a result code from an underlying API
...
 
 
+Warnings
+========
+
+.. py:class:: LDAPBytesWarning
+
+    Raised when bytes/text mismatch in non-strict bytes mode.
+
+    See :ref:`bytes_mode` for details.
```
tiran: needs ``versionadded``

PR #104, opened 2017-12-05
Support None for set_option(OPT_NETWORK_TIMEOUT)

tiran:
The setter for network timeout and timeout options now support None as
value. None is mapped to infinity.

Closes: https://github.com/python-ldap/python-ldap/issues/95

### DIFF
--- a/Modules/options.c
+++ b/Modules/options.c
@@ -136,20 +136,22 @@ LDAP_set_option(LDAPObject *self, int option, PyObject *value)
 	    break;
     case LDAP_OPT_TIMEOUT:
     case LDAP_OPT_NETWORK_TIMEOUT:
-        /* Float valued timeval options, 'd' also handles int/long */
-        if (!PyArg_Parse(value, "d:set_option", &doubleval)) {
-            /* clear error and try again with None */
-            PyErr_Clear();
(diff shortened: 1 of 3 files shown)


encukou:
Looks great at first glance!
I'll most likely get to reviewing/merging on Thursday.

encukou:
Rewriting the conversion error message should happen only on TypeError. I changed that, (hopefully) made the control flow a bit more straightforward, and removed the [nonstandard](https://docs.python.org/3/c-api/none.html) PyNone_Check.

Modules/options.c
```diff
@@ -136,20 +136,22 @@ LDAP_set_option(LDAPObject *self, int option, PyObject *value)
...
-                    Py_TYPE(value)->tp_name
-                );
+        /* Float valued timeval options */
+        if (value == Py_None) {
+            /* None is mapped to infinity timeout */
+            doubleval = -1;
+        } else {
+            /* 'd' handles int/long */
+            if (!PyArg_Parse(value, "d:set_option", &doubleval)) {
+                if (PyErr_ExceptionMatches(PyExc_TypeError)) {
```
tiran: Good catch! I haven't considered overflows.

Issue #105, opened 2017-12-06
Clean up memory handling in attrs_from_List()


tiran:
The function https://github.com/python-ldap/python-ldap/blob/084ffe03e14974b798d2fb913edbb586af7a0c94/Modules/LDAPObject.c#L251 has a rather peculiar way to handle memory. It takes a sequence of Python bytes / text and converts it to a NULL terminated array of ``char *``. It allocates memory for the array itself but uses ``PyUnicode_AsUTF8`` or ``PyBytes_AsString`` for the elements inside the array. Both functions do **not** return a copy of the string but rather a reference to an internal buffer. The internal buffer is shared between callers and must be treated as const, immutable block. In fact in Python 3.7+, the return value of ``PyUnicode_AsUTF8`` has been changed to ``const char*``, https://docs.python.org/3.7/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize . However ``ldap_search_ext`` does not take a ``const char*[]`` but a ``char *attrs[]``. This breaks the const promise starting with Python 3.7.

Even more peculiar is the ``seq`` argument. It is used to pass back a reference to the Python sequence that holds the ``PyObject *``. This trick is used to keep the Python sequence and all its members alive, so the elements in the array of ``char *`` do not get deallocated by Python.

I suggest to change the function to copy the strings to allocated buffers instead of (ab)using Python objects to provide the buffer. The performance impact of additional ``PyMem_NEW`` and ``memcpy`` is negligible.

PR #106, opened 2017-12-06
Change memory handling in attrs_from_List()

tiran:
attrs_from_List() no longer uses internal buffers of bytes/str PyObject*
to maintain memory. Instead it creates its own copies, which are then
passed to ldap_search_ext() and then cleaned up.

The modification is required for Python 3.7. PyUnicode_AsUTF8() returns
const char* but ldap_search_ext() does not use const.

Closes: https://github.com/python-ldap/python-ldap/issues/105

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -16,7 +16,7 @@
 #include <sasl/sasl.h>
 #endif
 
-static void free_attrs(char***, PyObject*);
+static void free_attrs(char***);
 
 /* constructor */
 
@@ -248,13 +248,10 @@ error:
  */
 
 int
-attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
+attrs_from_List( PyObject *attrlist, char***attrsp) {
 
     char **attrs = NULL;
-    Py_ssize_t i, len;
-    PyObject *item;
-
-    *seq = NULL;
+    PyObject *seq = NULL;
 
     if (attrlist == Py_None) {
         /* None means a NULL attrlist */
@@ -268,8 +265,16 @@ attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
             "expected *list* of strings, not a string", attrlist);
         goto error;
     } else {
-        *seq = PySequence_Fast(attrlist, "expected list of strings or None");
-        if (*seq == NULL)
+        PyObject *item = NULL;
+        Py_ssize_t i, len, strlen;
+#if PY_MAJOR_VERSION >= 3
+        const char *str;
+#else
+        char *str;
+#endif
+
+        seq = PySequence_Fast(attrlist, "expected list of strings or None");
+        if (seq == NULL)
             goto error;
 
         len = PySequence_Length(attrlist);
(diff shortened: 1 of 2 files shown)


Modules/LDAPObject.c
```diff
@@ -289,16 +294,27 @@ attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
                 LDAPerror_TypeError("expected bytes in list", item);
```
encukou: As I was reading this, I also changed the comment “Encoded by Python to UTF-8” to “Encoded *in* Python to UTF-8”, since it's python-ldap's Python code doing the encoding.

Modules/LDAPObject.c
```diff
@@ -268,8 +265,16 @@ attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
...
     } else {
-        *seq = PySequence_Fast(attrlist, "expected list of strings or None");
-        if (*seq == NULL)
+        PyObject *item = NULL;
+        Py_ssize_t i, len, strlen;
+#if PY_VERSION_HEX >= 0x03070000
+        const char *str;
+#else
+        char *str;
+#endif
```
encukou: It's should be fine to use `const char` everywhere.

encukou:
Rebased with the two changes above folded in.

Modules/LDAPObject.c
encukou: well, everywhere but Python 2.

Modules/LDAPObject.c
```diff
@@ -268,8 +265,16 @@ attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
             "expected *list* of strings, not a string", attrlist);
         goto error;
     } else {
-        *seq = PySequence_Fast(attrlist, "expected list of strings or None");
-        if (*seq == NULL)
+        PyObject *item = NULL;
+        Py_ssize_t i, len, strlen;
+#if PY_MAJOR_VERSION >= 3
```
tiran: That's wrong, it must be ``#if PY_VERSION_HEX >= 0x03070000``. Python < 3.7 has ``PyAPI_FUNC(char *) PyUnicode_AsUTF8AndSize()``. Only 3.7+ has ``PyAPI_FUNC(const char *) PyUnicode_AsUTF8AndSize()``

Modules/LDAPObject.c
```diff
@@ -280,25 +285,36 @@ attrs_from_List( PyObject *attrlist, char***attrsp, PyObject** seq) {
...
             if (!PyUnicode_Check(item)) {
                 LDAPerror_TypeError("expected string in list", item);
                 goto error;
             }
-            attrs[i] = PyUnicode_AsUTF8(item);
+            str = PyUnicode_AsUTF8AndSize(item, &strlen);
 #endif
+            /* Make a copy. PyBytes_AsString* / PyUnicode_AsUTF8* return
+             * internal values that must be treated like const char. Python
+             * 3.7 actually returns a const char.
```
tiran: See here

Modules/LDAPObject.c
encukou: I thought implicitly converting `char *` to `const char *` is perfectly fine – it just limits what can be done with the string later.
Why is `const char *str = PyUnicode_AsUTF8AndSize(...)` a problem on 3.6-?

tiran: I discussed the issue with @vstinner. He assured me that const is always fine here. Let's go for your version check.

tiran:
Your modifications LGTM

Issue #108, opened 2017-12-06
Warning 'Received non-bytes value' is not helpful


tiran:
The ``Received non-bytes value`` warning isn't helpful at all. The warning message only contains ``ldap/ldapobject.py`` as caller information, not the actual caller. It might be enough to increment the stack level of the warning by one.

Example:

``/usr/lib64/python2.7/site-packages/ldap/ldapobject.py:396: BytesWarning: Received non-bytes value u'givenname' with default (disabled) bytes mode; please choose an explicit option for bytes_mode on your LDAP connection``

tiran:
PR #109 is a proof of concept to use correct stack levels. python-ldap uses up to three, maybe for stack levels for methods, so it's getting ugly. :/

PR #109, opened 2017-12-06
[PoC] Fix stacklevel for warning

tiran:
Closes: https://github.com/python-ldap/python-ldap/issues/108


tiran:
The approach is too complicated, let's do #128

PR #113, opened 2017-12-08
Provide build deps for Alpine and CentOS

tiran:
Also updates build devs for Fedora and Debian.

See: https://github.com/pyldap/pyldap/issues/91#issuecomment-350098897


encukou:
Rebased & improved formatting a bit

PR #117, opened 2017-12-12
Prefer literals throughout project

jdufresne:
- Use a dict literal instead of calling `dict()`.
- Use a list literal instead of calling `list()`.
- Use a set literal instead of calling `set()`.
- Use a tuple literal instead of calling `tuple()`.
- Use dict comprehension

These syntax expressions are available by all supported Python versions. Literals are always faster than calling the constructor. More idiomatic with wider Python community.


tiran:
Thanks, LGTM

PR #118, opened 2017-12-12
Trim trailing whitespace throughout project

jdufresne:
- Trim whitespace from end of lines
- Trim whitespace from end of files

Helps avoid spurious changes during pull requests, which leads to cleaner diffs during code reviews.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,12 +12,12 @@ Issue description:
 
 
 
-Steps to reproduce: 
+Steps to reproduce:
 
 
 
-Operating system: 
+Operating system:
 
-Python version: 
+Python version:
 
-python-ldap version: 
+python-ldap version:
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-
 # Auto-generated
 .*.swp
 *.pyc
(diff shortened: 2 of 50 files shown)


tiran:
Thanks for your contribution. We are planning to to a full code reformatting of C code and Python code eventually, see PR #2 . Could you please add a comment to the PR and post the command you have used?

encukou:
Thanks for the suggestion!

As @tiran said, we're delaying *wholesale* code reformatting – solely because there are/were a lot of in-progress patches that would need rebasing.
Now, this particular PR doesn't look hard to rebase onto, so I wouldn't mind merging it if you *really* want it in for 3.0. After 3.0, we'll run autopep8 to fix all the whitespace. (I don't think an extra command will be needed.)

I'll close the PR, but feel free to re-open. Avoid changing non-code files though: `.github/ISSUE_TEMPLATE.md` has the spaces on purpose, and PEM certificates are generated data.

jdufresne:
>  Could you please add a comment to the PR and post the command you have used?

The command is:

```
$ find . -name .git -prune -o -type f -exec emacs -q -batch {} -eval "(require 'whitespace)" -eval "(let ((whitespace-style '(empty trailing))) (whitespace-cleanup))" -f save-buffer \; 
```

With some slight post processing to revert auto-generated files.

PR #120, opened 2017-12-12
Auto-detect stacklevel for bytes warning

tiran:
Alternative approach to PR #109. I'm walking up the stack frames to find the first frame that is outside the python-ldap's ldapobject module.

Closes: https://github.com/python-ldap/python-ldap/issues/108


encukou:
Fixed in https://github.com/python-ldap/python-ldap/pull/128

PR #122, opened 2017-12-12
Run Travis tests with PyPy

tiran:
To my surprose, python-ldap works just fine with PyPy2. For Python 3,
PyPy-5.9 (Python 3.5) is required. Fedora's and Ubuntu's PyPy 5.5 is
too old (Python 3.3).


encukou:
I'm also quite surprised this just works. PyPy devs have (even more of) my respect.

PR #123, opened 2017-12-12
Drop distutils workaround and simplify setup.py

jdufresne:
Modern Python installations contain setuptools. It is reasonable to expect it to exist. Simplifies `setup.py` and removes the large comment explaining the workaround.


encukou:
Thanks, I like the direction!
I'll do a review tomorrow.

encukou:
Looks good, thank you!

PR #124, opened 2017-12-13
Remove long deprecated functions ldap.open() and ldap.init()

jdufresne:
The functions have been deprecated since at least the introduction of the file `Lib/ldap/functions.py` in commit e4173acd69254b950dfeb7663e83a8e338f5df18 on 2011-11-23. That has been lots of time for library users to see the warning and update calling code.

With the move to 3.0.0, a new major version, this is an opportune time to drop deprecated, unused functionality.

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -66,28 +66,6 @@ This module defines the following functions:
 
       :rfc:`4516` - Lightweight Directory Access Protocol (LDAP): Uniform Resource Locator
 
-.. py:function:: open(host [, port=PORT]) -> LDAPObject object
-
-   Opens a new connection with an LDAP server, and return an LDAP object  (see
-   :ref:`ldap-objects`) used to perform operations on that server.  *host* is a
-   string containing solely the host name. *port* is an integer specifying the
-   port where the LDAP server is  listening (default is 389).
-
-   .. deprecated:: 3.0
-
-      ``ldap.open()`` is deprecated. It will be removed in version 3.1. Use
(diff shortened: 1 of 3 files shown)


tiran:
+1 That's a great idea. Thanks for taking care of the deprecated functions.

It might be useful to include a change information in the documentation. What do you think about something like this at the end of the function section?

```
.. versionchanged: 3.0
   The deprecated function `init` and `open` were removed.
```

I also noticed that ``ldap.initialize`` does not contain examples for LDAP URIs. An ldap, ldaps and ldapi URI plus link to ``ldapurl`` documentation might be useful.

tiran:
What do you think about a ``.. deprecated:`` note in 3.0 docs?

jdufresne:
> It might be useful to include a change information in the documentation. What do you think about something like this at the end of the function section?

Done in the latest revision.

> I also noticed that ldap.initialize does not contain examples for LDAP URIs. An ldap, ldaps and ldapi URI plus link to ldapurl documentation might be useful.

This seems like a separate concern than what is being addressed by this work. The deleted documentation also did not have this information. I think this would be fine as a separate PR.

jdufresne:
> Let's do this in 3.1. We're in the final stretch before 3.0; the second beta is not a good time to remove features.

Makes sense.

I only learned about the new GitHub home for python-ldap recently. Waiting for 3.1 sounds fine to me.

jdufresne:
> What do you think about a .. deprecated: note in 3.0 docs?

Done in #126

PR #128, opened 2017-12-13
When raising LDAPBytesWarning, walk the stack to determine stacklevel

encukou:
This is a simplification of https://github.com/python-ldap/python-ldap/pull/120 (and https://github.com/python-ldap/python-ldap/pull/109).

Turns out extension functions have `globals()`, and putting the marker in the `_ldap` module as well means we can only walk the stack – no other changes necessary.
BTW I do like removing `_bytesify_inputs`! Let's do that in 3.1.

I did have some trouble in py2 tests, see the `_normalize` inner function.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -39,6 +39,9 @@ else:
   text_type = str
 
 
+# See SimpleLDAPObject._bytesify_input
+_LDAP_WARN_SKIP_FRAME = True
+
 class LDAPBytesWarning(BytesWarning):
   """python-ldap bytes mode warning
   """
@@ -126,11 +129,27 @@ class SimpleLDAPObject:
         if self.bytes_mode_hardfail:
           raise TypeError("All provided fields *must* be bytes when bytes mode is on; got %r" % (value,))
         else:
+          # Raise LDAPBytesWarning.
+          # Call stacks with _bytesify_input tend to be complicated, so
+          # getting a useful stacklevel is tricky.
+          # We walk stack frames, ignoring all functions in this file
+          # and in the _ldap extension, based on a marker in globals().
+          stacklevel = 0
+          try:
+            getframe = sys._getframe
+          except AttributeError:
+            pass
+          else:
+            frame = sys._getframe(stacklevel)
+            # walk up the stacks until we leave the file
+            while frame and frame.f_globals.get('_LDAP_WARN_SKIP_FRAME'):
+                stacklevel += 1
+                frame = frame.f_back
           warnings.warn(
             "Received non-bytes value %r with default (disabled) bytes mode; please choose an explicit "
             "option for bytes_mode on your LDAP connection" % (value,),
             LDAPBytesWarning,
-            stacklevel=6,
+            stacklevel=stacklevel+1,
           )
           return value.encode('utf-8')
     else:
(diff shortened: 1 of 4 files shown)


encukou:
@tiran, does this look OK?

Lib/ldap/ldapobject.py
```diff
@@ -126,11 +129,27 @@ def _bytesify_input(self, value):
...
           raise TypeError("All provided fields *must* be bytes when bytes mode is on; got %r" % (value,))
         else:
+          # Raise LDAPBytesWarning.
+          # Call stacks with _bytesify_input tend to be complicated, so
+          # getting a useful stacklevel is tricky.
+          # We walk stack frames, ignoring all functions in this file
+          # and in the _ldap extension, based on a marker in globals().
+          stacklevel = 0
+          try:
+            getframe = frame = sys._getframe
```
tiran: Why assign to ``frame``?

Lib/ldap/ldapobject.py
```diff
@@ -126,11 +129,27 @@ def _bytesify_input(self, value):
         if self.bytes_mode_hardfail:
           raise TypeError("All provided fields *must* be bytes when bytes mode is on; got %r" % (value,))
         else:
+          # Raise LDAPBytesWarning.
+          # Call stacks with _bytesify_input tend to be complicated, so
+          # getting a useful stacklevel is tricky.
+          # We walk stack frames, ignoring all functions in this file
+          # and in the _ldap extension, based on a marker in globals().
+          stacklevel = 0
```
tiran: You can start with an offset of 2, maybe 3.

encukou:
Thanks, will fix.
The same thing needs to be done for the warning in `__init__`; `ReconnectLDAPObject` calls its superclass there.
I'm testing an update.

encukou:
this is slowly leaving "simplification" territory...

Lib/ldap/ldapobject.py
```diff
@@ -39,10 +39,34 @@
...
+    # based on the _LDAP_WARN_SKIP_FRAME marker in globals().
+    stacklevel = 2
+    try:
+        getframe = sys._getframe
+    except AttributeError:
+        pass
+    else:
+        frame = sys._getframe(stacklevel)
+    # walk up the stacks until we leave the file
+    while frame and frame.f_globals.get('_LDAP_WARN_SKIP_FRAME'):
```
tiran: frame is not defined in case ``sys._getframe`` does not exist. Can we rather wrap the whole block into a ``hasattr(sys, '_getframe')`` check?

encukou: Oops, thanks for the catch!

encukou:
I'll be away until Friday, and I'd like to release beta 3 on Friday or Monday.
@tiran, feel free to take over if there's anything more to be done here.

tiran:
The latest version of the PR addresses the stacklevel issue. I'm not quite happy with warnings in general. There are two potential issues:

1) It's still hard to pin point bytes issues. The warning message does not include a hint, which argument is wrong. We may want to consider to include the name of the argument (e.g. ``dn``) which is wrong.
2) I suspect (without profiling) that warnings are a performance regression for ``add*`` and ``search*``. All add op take a ``modlist``, all search ops take an optional ``attrlist``. The lists can be pretty long and we do the full warning dance for every entry. We should only warn once per operation.

Both issues can be addressed in new PRs.

PR #131, opened 2017-12-13
Add tests and coverage for tracing

tiran:
python-ldap can trace LDAP calls, but it was not possible to use the
feature without changing code manually. Tracing can now be enabled with
the env var PYTHON_LDAP_TRACE_LEVEL and redirected to a file with
PYTHON_LDAP_TRACE_FILE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,9 @@ matrix:
         - TOXENV=py3-nosasltls
         - WITH_GCOV=1
     - python: 3.6
+      env:
+        - TOXENV=py3-trace
+    - python: 3.6
       env: TOXENV=doc
   allow_failures:
      - env:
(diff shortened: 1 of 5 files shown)


tiran:
Rather than adding tons of ``pragma: no cover``, let's test the trace feature.

encukou:
The new env vars need to be documented.
I'd rather put this in 3.1, unless there's some good reason to rush.

I'm not sure the pickling the trace level, but not the file, is a good idea. I think it would make sense to stop pickling these, and always use the global settings. What do you think?

tiran:
3.1 is fine

I haven't documented the env vars because I'm not sure that it should be a publicly available, supported API.

encukou:
IMO we should either document it, or get rid of it.
If it's not public API, it at least needs a quick mention in developer docs.

encukou:
Rebased. I also added the new tox env to `.travis.yml`, so it actually gets run. Let's see if it works.

PR #137, opened 2017-12-18
Drop use of .has_key(); deprecated in modern Python

jdufresne:
Using the `in` operator is preferred and available on all supported versions of Python.

From https://docs.python.org/3/whatsnew/3.0.html#builtins

> Removed. `dict.has_key()` – use the `in` operator instead.

### DIFF
--- a/Lib/ldap/modlist.py
+++ b/Lib/ldap/modlist.py
@@ -73,7 +73,6 @@ def modifyModlist(
       replace_attr_value = len(old_value)!=len(new_value)
       if not replace_attr_value:
         if attrtype_lower in case_ignore_attr_types:
-          norm_func = str.lower
           old_value_set = set(map(str.lower,old_value))
           new_value_set = set(map(str.lower,new_value))
         else:


encukou:
Indeed. On Python 2, `has_key` needs to stay.

jdufresne:
Should I add a warning for its eventual removal?

Python 2.7 users can continue to use the `in` operator. It is supported.

encukou:
> Should I add a warning for its eventual removal?

Yes, a DeprecationWarning sounds good. But I wouldn't rush the removal itself. Let's say we'll remove it when Python 2 support is dropped?

tiran:
We could only add the function under Python 2 and not define it under Python 3.

encukou:
> We could only add the function under Python 2 and not define it under Python 3.

We are past the stage of defining an API for python-ldap 3.0.

Sure, `has_key` looks ugly and backwards and "the wrong way to do it", and a good linter will complain you about it... but it's there, it's tested, it's not a source of subtle bugs, and it doesn't really cost anything to keep it.
Let's make it clear it's not the way to do things, but avoid breaking people's code for purity's sake.

jdufresne:
Updated. Restored function. Now issues a warning.

encukou:
Hi @jdufresne,
Embarrassingly, when I was rebasing your other pull request to get rid of conflicts, I managed to overwrite your branch here.
Could you please push your changes again?
I'm sorry for this. I'll be more careful in the future.

jdufresne:
No worries. Done in #151 Thanks

encukou:
That's not the one. Could you push your local branch `drop-has-key` to GitHub?

jdufresne:
Oops, sorry about that. Lets try that again, #153

encukou:
Thank you! The change looks good; I'll do a full review next week.

PR #138, opened 2017-12-18
Prefer iterating dict instead of calling dict.keys()

jdufresne:
Calling `dict.keys()` is unnecessary. `iter(dict)` is equivalent to `dict.keys()`. Inspired by Lennart Regebro's talk "Prehistoric Patterns in Python" from PyCon 2017.

https://www.youtube.com/watch?v=V5-JH23Vk0I

### DIFF
--- a/Lib/ldap/modlist.py
+++ b/Lib/ldap/modlist.py
@@ -50,7 +50,7 @@ def modifyModlist(
   case_ignore_attr_types = {v.lower() for v in case_ignore_attr_types or []}
   modlist = []
   attrtype_lower_map = {}
-  for a in old_entry.keys():
+  for a in old_entry:
     attrtype_lower_map[a.lower()]=a
   for attrtype, value in new_entry.items():
     attrtype_lower = attrtype.lower()
--- a/Lib/ldap/schema/subentry.py
+++ b/Lib/ldap/schema/subentry.py
@@ -23,7 +23,7 @@ for o in list(vars().values()):
     SCHEMA_CLASS_MAPPING[o.schema_attribute] = o
     SCHEMA_ATTR_MAPPING[o] = o.schema_attribute
 
-SCHEMA_ATTRS = SCHEMA_CLASS_MAPPING.keys()
+SCHEMA_ATTRS = list(SCHEMA_CLASS_MAPPING)
 
 
 class SubschemaError(ValueError):
@@ -122,7 +122,7 @@ class SubSchema:
         self.sed[se_class][se_id] = se_instance
 
         if hasattr(se_instance,'names'):
-          for name in ldap.cidict.cidict({}.fromkeys(se_instance.names)).keys():
+          for name in ldap.cidict.cidict({}.fromkeys(se_instance.names)):
             if check_uniqueness and name in self.name2oid[se_class]:
               self.non_unique_names[se_class][se_id] = None
               raise NameNotUnique(attr_value)
@@ -130,7 +130,7 @@ class SubSchema:
               self.name2oid[se_class][name] = se_id
 
     # Turn dict into list maybe more handy for applications
-    self.non_unique_oids = self.non_unique_oids.keys()
+    self.non_unique_oids = list(self.non_unique_oids)
 
     return # subSchema.__init__()
 
@@ -168,7 +168,7 @@ class SubSchema:
           except AttributeError:
             pass
     else:
-      result = avail_se.keys()
+      result = list(avail_se)
     return result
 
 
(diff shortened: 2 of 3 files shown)


Lib/ldap/cidict.py
```diff
@@ -35,7 +35,7 @@ def __delitem__(self,key):
     del self.data[lower_key]
 
   def update(self,dict):
-    for key in dict.keys():
+    for key in dict:
```
tiran: You can drop the whole function. ``UserDict.IterableUserDict`` and ``collections.Mapping`` always provide a correct ``update()``.

Lib/ldap/cidict.py
```diff
@@ -50,6 +50,9 @@ def get(self,key,failobj=None):
     except KeyError:
       return failobj
 
+  def __iter__(self):
```
tiran: ``UserDict.IterableUserDict`` and ``collections.Mapping`` always provide ``__iter__``

Lib/ldap/schema/models.py
```diff
@@ -655,7 +655,7 @@ def _at2key(self,nameoroid):
       return t
 
   def update(self,dict):
-    for key in dict.keys():
+    for key in dict:
```
tiran: Same here

Lib/ldap/cidict.py
jdufresne: True, but it is a different implementation. 

https://github.com/python/cpython/blob/2.7/Lib/UserDict.py#L113-L114

The Python one iterates over the keys of stored data (`self.data.keys()`). For cidict, this will be the lowercased versions. But for cidict, we should use the original vales.

I could drop this `__iter__` if the library were changed to:

```py
  def __setitem__(self,key,value):
    lower_key = key.lower()
    self._keys[key] = lower_key
    self.data[key] = value

  def __getitem__(self,key):
    key = self._keys.get(key.lower(), key)
    return self.data[key.lower()]
```

Otherwise, this method will need to stay. What do you think?

Lib/ldap/cidict.py
jdufresne: Using the parent implementation results in not correctly transforming the keys to lowercase on Python 2.

See an experimental test case here:

https://travis-ci.org/jdufresne/python-ldap/builds/318122299

jdufresne:
> However dict.keys() to dict iteration can cause dict size change during iteration bugs. I'd prefer to delay this postpone this change for 3.1 release.

This would already be true on Python 3. On Python 3, you can't both iterate over `dict.keys()` and modify the dict as `dict.keys()` is just a view and not a list copy of the keys.

encukou:
@jdufresne, can I rebase this? (It's not a problem for me, but I don't want to cause conflicts if you're working on it.)

jdufresne:
Sure, go ahead, thanks. I normally just grep for `\.keys(` and analyze the results.

jdufresne:
Rebased on the latest master branch. Thanks.

Demo/matchedvalues.py
```diff
@@ -35,7 +35,7 @@
 def print_result(search_result):
     for n in range(len(search_result)):
         print("dn: %s" % search_result[n][0])
-        for attr in search_result[n][1].keys():
+        for attr in search_result[n][1]:
```
encukou: This isn't covered in the test suite. I'd prefer not changing it until it has tests.

Lib/ldap/schema/models.py
```diff
@@ -655,7 +655,7 @@ def _at2key(self,nameoroid):
       return t
 
   def update(self,dict):
-    for key, value in dict.values():
+    for key, value in dict:
```
encukou: This one is wrong.

Lib/ldapurl.py
```diff
@@ -130,14 +130,14 @@ def __ne__(self,other):
     return not self.__eq__(other)
 
 
-class LDAPUrlExtensions(UserDict):
+class LDAPUrlExtensions(IterableUserDict):
```
encukou: I'd rather use `MutableMapping` than `IterableUserDict`, for reasons I give in https://github.com/python-ldap/python-ldap/pull/186.

encukou:
Sorry for the delay in reviewing.
If you'd like, I can rebase this, address my comments and let you review it. It would make the back-and-forth faster, but I don't want to mess with your pull request without warning.

Lib/ldapurl.py
jdufresne: I've reverted this change to defer it to the linked PR.

Lib/ldap/schema/models.py
jdufresne: Thanks for catching. I've reverted it.

Demo/matchedvalues.py
jdufresne: No problem. I've reverted it.

jdufresne:
>  If you'd like, I can rebase this, address my comments and let you review it. It would make the back-and-forth faster, but I don't want to mess with your pull request without warning.

That is fine with me. Please feel free to make additional changes should a review catch anything else.

PR #140, opened 2017-12-18
Remove unused local variable, norm_func

jdufresne:


jdufresne:
> How about old_value_set = set(s.lower() for s in old_value) instead of map?

That could even use set comprehension:

```py
{s.lower() for s in old_value}
```

If you'd like, I'll begin working on a PR to convert the uses of `map()` to use list, dict, or set comprehension instead. I'll do this as separate work though.

jdufresne:
> How about old_value_set = set(s.lower() for s in old_value) instead of map?

Done in #142

PR #141, opened 2017-12-18
Make testing on non-Linux platforms easier

tiran:
A bunch of small chances to the test suite. It should be enough to make it possible to run the test suite under macOS homebrew and Windows (untested).

* Make LDAPI optional
* Drop unused import of pwd module
* Lookup test binaries in PATH

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -56,6 +56,12 @@ authz-regexp
 
 LOCALHOST = '127.0.0.1'
 
+CI_DISABLED = set(os.environ.get('CI_DISABLED', '').split(':'))
+if 'LDAPI' in CI_DISABLED:
+    HAVE_LDAPI = False
+else:
+    HAVE_LDAPI = hasattr(socket, 'AF_UNIX')
+
 
 def identity(test_item):
     """Identity decorator
@@ -69,7 +75,7 @@ def skip_unless_ci(reason, feature=None):
     """
     if not os.environ.get('CI', False):
         return unittest.skip(reason)
-    elif feature in os.environ.get('CI_DISABLED', '').split(':'):
+    elif feature in CI_DISABLED:
         return unittest.skip(reason)
     else:
         # Don't skip on Travis
@@ -91,6 +97,14 @@ def requires_sasl():
     if not ldap.SASL_AVAIL:
         return skip_unless_ci(
             "test needs ldap.SASL_AVAIL", feature='SASL')
+    else:
+        return identity
+
+
+def requires_ldapi():
+    if not HAVE_LDAPI:
+        return skip_unless_ci(
+            "test needs ldapi support (AF_UNIX)", feature='LDAPI')
     else:
         return identity
 
@@ -149,8 +163,6 @@ class SlapdObject(object):
     root_dn = 'cn=%s,%s' % (root_cn, suffix)
     root_pw = 'password'
     slapd_loglevel = 'stats stats2'
-    # use SASL/EXTERNAL via LDAPI when invoking OpenLDAP CLI tools
-    cli_sasl_external = True
     local_host = '127.0.0.1'
     testrunsubdirs = (
         'schema',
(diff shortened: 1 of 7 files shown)


Lib/slapdtest/_slapdtest.py
```diff
@@ -56,6 +56,12 @@
 
 LOCALHOST = '127.0.0.1'
 
+CI_DISABLED = set(os.environ.get('CI_DISABLED', '').split(':'))
+if 'LDAPI' in CI_DISABLED:
+    HAVE_LDAPI = False
+else:
+    HAVE_LDAPI = hasattr(socket, 'AF_UNIX')
+
```
encukou: What if, when `CI` is set, we only looked at `CI_DISABLED`? Maybe with an early error if `LDAPI` isn't disabled but `socket.AF_UNIX` is available.
And if `CI` is unset, do the `hasattr` check.

In this case, the `hasattr` is probably robust enough, but I think starting a "avoid feature detection if `CI` is set" pattern would be beneficial.

Lib/slapdtest/_slapdtest.py
```diff
@@ -331,11 +352,14 @@ def _start_slapd(self):
...
         """
+        urls = [self.ldap_uri]
+        if self.ldapi_uri:
+            urls.append(self.ldapi_uri)
         slapd_args = [
             self.PATH_SLAPD,
             '-f', self._slapd_conf,
             '-F', self.testrundir,
-            '-h', '%s' % ' '.join((self.ldap_uri, self.ldapi_uri)),
+            '-h', '%s' % ' '.join(urls),
```
encukou: The `%` interpolation looks unnecessary here.

Lib/slapdtest/_slapdtest.py
```diff
@@ -109,6 +110,35 @@ def requires_ldapi():
...
+        files = [cmd]
+
+    for directory in path:
+        for name in files:
+            name = os.path.join(directory, name)
+            if os.path.exists(name) and os.access(name, os.F_OK | os.X_OK):
+                return name
+    return None
+
+
```
encukou: Here too, I feel it would be nice to avoid auto-discovery if `CI` is set.

Lib/slapdtest/_slapdtest.py
```diff
@@ -223,13 +251,15 @@ def __init__(self):
...
-                raise ValueError('Binary {} is missing.'.format(binary))
+        for name in dir(self):
+            if not name.startswith('PATH_'):
+                continue
+            value = getattr(self, name)
+            if value is None or not os.path.isfile(value):
+                cmd = name[5:].lower()
+                raise ValueError(
+                    "Command '{}' not found in PATH".format(cmd)
+                )
```
encukou: So … commands are defined by being prefixed by `PATH_`?
That looks like avoidable magic. I think it would be better to have the explicit list and just add a comment to remind people about synchronization, or to turn `_which` into something that registers the command as well as finding it.

encukou:
Should we turn on macOS on Travis, as well?

Lib/slapdtest/_slapdtest.py
tiran: I'd like to run tests without LDAPI support locally, too. Can you recommend a better way to pass an option from tox without introducing more env vars?

We could move to pytest in order to use proper fixtures and options. But I don't like to do that now.

Lib/slapdtest/_slapdtest.py
tiran: Good catch

Lib/slapdtest/_slapdtest.py
tiran: What do you mean by auto-detection here? ``_which`` simply looks through PATH.

Lib/slapdtest/_slapdtest.py
tiran: The old list of file names doesn't result in good UX because ``_which`` returns ``None`` when it can't find a command. An error message like ``command None not found`` isn't very helpful. What do you suggest?

Lib/slapdtest/_slapdtest.py
encukou: I misunderstood the code. Sorry for the noise!

encukou:
Here is my suggestion: do both the `which()` and the error handling from `__init__`, rather than at the class level and from `start()`. This should be fine: we don't really care about the performance hit of doing it finding paths for each instance, and the whole class is useless if `start()` fails, so no harm in raising errors already in `__init__`.

I also made `which()` clearly a backport, not a modified fork.

tiran:
+1 for which()

But I don't like the sbin code. It's too complicated. Should we really care where a command is located as long as it's found in a standard bin dir? I also don't like SLAPD, BINDIR and SBINDIR env vars. They are hacks that are no longer required with which.

encukou:
I agree they're too complicated. However, since 2.4.36, slapdtest is a top-level module for general use, rather than an internal python-ldap helper. I consider the env vars to be its public API. So, I don't want to drop them in 3.0.

It might make sense to deprecate slapdtest as a general tool, and later make it an internal test helper again.

tiran:
I'm not fully satisfied with the patch, but it's as good as it will get. But it should do the trick while keeping backwards compatibility. Let's ship it!

PR #142, opened 2017-12-19
Update uses of map() to use list/set comprehensions instead

jdufresne:
Updates style to modern Python idioms.

### DIFF
--- a/Lib/ldap/controls/deref.py
+++ b/Lib/ldap/controls/deref.py
@@ -108,7 +108,7 @@ class DereferenceControl(LDAPControl):
     for deref_res in decodedValue:
       deref_attr,deref_val,deref_vals = deref_res[0],deref_res[1],deref_res[2]
       partial_attrs_dict = {
-        str(tv[0]): map(str,tv[1])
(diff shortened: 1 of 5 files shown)


Lib/ldap/controls/deref.py
```diff
@@ -108,7 +108,7 @@ def decodeControlValue(self,encodedControlValue):
     for deref_res in decodedValue:
       deref_attr,deref_val,deref_vals = deref_res[0],deref_res[1],deref_res[2]
       partial_attrs_dict = {
-        str(tv[0]): map(str,tv[1])
+        str(tv[0]): [str(v) for v in tv[1]]
```
tiran: In Python 2, list comprehensions leak the internal variable. Therefore I prefer list() for list comprehensions.

encukou: I think it's OK to use comprehensions with one-letter variable names.
No need to change this.

encukou:
3.0 will need another change, so let's include this one in 3.0 as well.

Issue #144, opened 2017-12-19
Password disclosure in _raise_byteswarning()


tiran:
``_raise_byteswarning()`` helper function now includes the value in the warning message. This may disclose sensitive information like passwords or attributes in server logs.

Issue #132, opened 2017-12-15
Make bytes warnings more useful


tiran:
From https://github.com/python-ldap/python-ldap/pull/128#issuecomment-351941287

It's still hard to pin point bytes issues. The warning message does not include a hint, which argument is wrong. We may want to consider to include the name of the argument with invalid type (e.g. ``dn``) in the warning message. python-ldap must not include the actual value, because it may contain confidential information like credentials.

PR #145, opened 2017-12-19
Show argument name instead of content in LDAPBytesWarning

encukou:
Based on earlier patch by Christian Heimes
Fixes: https://github.com/python-ldap/python-ldap/issues/132
Fixes: https://github.com/python-ldap/python-ldap/issues/144

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -122,7 +122,7 @@ class SimpleLDAPObject:
     # On by default on Py2, off on Py3.
     self.bytes_mode = bytes_mode
 
-  def _bytesify_input(self, value):
+  def _bytesify_input(self, arg_name, value):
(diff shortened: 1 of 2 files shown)


Lib/ldap/ldapobject.py
```diff
@@ -532,7 +522,7 @@ def delete_ext(self,dn,serverctrls=None,clientctrls=None):
         form returns the message id of the initiated request, and the
         result can be obtained from a subsequent call to result().
     """
-    dn = self._bytesify_input(dn)
+    dn = self._bytesify_input('dn', dn)
```
tiran: ``if PY2:``?

encukou: `_bytesify_input` already contains a check. I only added the `PY2` check when there were several calls to `_bytesify_input`.

PR #148, opened 2017-12-29
Document all_records attribute of class LDIFRecordList

arugifa:
The current documentation of this attribute was not displayed on the online doc.


encukou:
Thank you!
Also, happy New Year – and apologies for the associated delay in reviewing!

Issue #147, opened 2017-12-21
Some default arguments are not compatible with bytes_mode=True


tiran:
With ``bytes_mode=True``, all arguments must be bytes. However some functions have default keyword arguments of type str (unicode). One example is ``search_ext(..., filterstr='(objectClass=*), ...')``.
  
## function signature

- [x] ``simple_bind(self,who='',cred='',serverctrls=None,clientctrls=None)``
- [x] ``simple_bind_s(self,who='',cred='',serverctrls=None,clientctrls=None)``
- [x] ``search_ext(self,base,scope,filterstr='(objectClass=*)', ...)``
- [x] ``search_ext_s(self,base,scope,filterstr='(objectClass=*)', ...)``
- [x] ``search(self,base,scope,filterstr='(objectClass=*)', ...)``
- [x] ``search_s(self,base,scope,filterstr='(objectClass=*)', ...)``
- [x] ``search_st(self,base,scope,filterstr='(objectClass=*)', ...)``
- [x] ``search_subschemasubentry_s(self,dn='')``
- [x] ``find_unique_entry(self,base,scope=ldap.SCOPE_SUBTREE,filterstr='(objectClass=*)', ...)``
- [x] ``read_rootdse_s(self, filterstr='(objectClass=*)', attrlist=None)``

## function body

- [x] ``search_subschemasubentry_s``
- [x] ``read_subschemasubentry_s``
- [x] ``find_unique_entry``
- [x] ``read_rootdse_s``
- [x] ``get_naming_contexts``
- [x] ``ReconnectLDAPObject._apply_last_bind``  

## After fixing this
- [ ] Remove `filterstr` workaround in `Tests/t_ldap_syncrepl.py`

anazmy:
Could this be related ?
~~~
Dec 31 05:46:28 server systemd: Starting IPA key daemon...
Dec 31 05:46:29 server ipa-dnskeysyncd: ipa         : INFO     LDAP bind...
Dec 31 05:46:29 server ipa-dnskeysyncd: ipa         : INFO     Commencing sync process
Dec 31 05:46:29 server ipa-dnskeysyncd: Traceback (most recent call last):
Dec 31 05:46:29 server ipa-dnskeysyncd: File "/usr/libexec/ipa/ipa-dnskeysyncd", line 114, in <module>
Dec 31 05:46:29 server ipa-dnskeysyncd: while ldap_connection.syncrepl_poll(all=1, msgid=ldap_search):
[...]

PR #150, opened 2018-01-03
Refactor syncrepl tests to run with bytes_mode

tiran:


encukou:
Looks great, thanks!
The filterstr is weird, but that's tracked in https://github.com/python-ldap/python-ldap/issues/147

(And I'd prefer formatting changes in their own commit, but I know it's a pain to separate them.)

PR #151, opened 2018-01-05
Prefer iterating dict instead of calling dict.keys()

jdufresne:
Calling `dict.keys()` is unnecessary. `iter(dict)` is equivalent to `dict.keys()`. Inspired by Lennart Regebro's talk "Prehistoric Patterns in Python" from PyCon 2017.

https://www.youtube.com/watch?v=V5-JH23Vk0I


encukou:
Duplicate of https://github.com/python-ldap/python-ldap/pull/138

PR #152, opened 2018-01-05
More bytesmode

tiran:
See https://github.com/python-ldap/python-ldap/issues/147

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -1070,7 +1070,7 @@ and wait for and return with the server's result, or with
 
    .. versionchanged:: 3.0
 
-      ``filterstr=None`` is equal to ``filterstr='(objectClass=*)'``.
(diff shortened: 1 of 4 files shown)


encukou:
It's a hack, but looks reasonable. I was considering doing something similar :)

I'll do a review next week.

Doc/reference/ldap.rst
```diff
@@ -1068,7 +1068,10 @@ and wait for and return with the server's result, or with
    or :py:meth:`search_ext_s()` (client-side search limit). If non-zero
    not more than *sizelimit* results are returned by the server.
 
+   .. versionchanged:: 3.0
 
+      ``filterstr=None`` is equivalent to ``filterstr='(objectClass=*)'``.
+
```
encukou: Looks good; the only nitpick I found is to say `equivalent` instead of `equal` here.

PR #154, opened 2018-01-09
Remove support for EOL Python 3.3

jdufresne:
Python 3.3 is end of life. It is no longer receiving bug fixes including
for security issues. It has been EOL since 2017-09-29. For a library
focused on authentication, supporting environments that are not
receiving security sends the wrong message. For additional details, see:

https://devguide.python.org/#status-of-python-branches

Reduces testing and maintenance resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,6 @@ matrix:
       env:
         - TOXENV=py27
         - WITH_GCOV=1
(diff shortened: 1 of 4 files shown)


encukou:
The coverage report here is suspicious, but let's put this in 3.0.

Issue #156, opened 2018-01-10
Valgrind finds memory leak in ldap_set_option


encukou:
Running `make valgrind` gives me a leak notice:
```
==22737== 27 bytes in 1 blocks are definitely lost in loss record 71 of 8,573
==22737==    at 0x4C2FB6B: malloc (vg_replace_malloc.c:299)
==22737==    by 0x171CC6EE: PL_strdup (in /usr/lib64/libplc4.so)
==22737==    by 0x15D9AF7B: ??? (in /usr/lib64/libldap_r-2.4.so.2.10.8)
==22737==    by 0x15D9516C: ??? (in /usr/lib64/libldap_r-2.4.so.2.10.8)
==22737==    by 0x15D8B3CC: ldap_set_option (in /usr/lib64/libldap_r-2.4.so.2.10.8)
==22737==    by 0x15B5489C: LDAP_set_option (options.c:189)
==22737==    by 0x15B4DBC5: l_ldap_set_option (LDAPObject.c:1295)
==22737==    by 0x4FC21AE: _PyCFunction_FastCallDict (in /usr/lib64/libpython3.6m.so.1.0)
==22737==    by 0x4FC9279: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==22737==    by 0x4FF2EC9: _PyEval_EvalFrameDefault (in /usr/lib64/libpython3.6m.so.1.0)
==22737==    by 0x4F8BBE9: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==22737==    by 0x4FC933D: ??? (in /usr/lib64/libpython3.6m.so.1.0)
==22737== 
```

Fedora 27, Python 3.6, python-ldap master (3.0.0b4)

tiran:
It's a known one-time memory leak in libldap's NSS engine. The leak is ignored by a rule. That rule does not match for you because you don't have debug symbols installed. I'll update the rule with a better one.

```
{
   NSS backend leaks one string during first initialization
   Memcheck:Leak
   match-leak-kinds: definite
   fun:malloc
   fun:PL_strdup
   fun:tlsm_init
   ...
}
```

encukou:
Argh. I see.

PR #157, opened 2018-01-10
Improve valgrind rule for PR_strdup

tiran:
libldap's NSS engine has one known memory leak. One block of memory is
leaked one time only. The existing rule only matches when all debug
symbols are installed. The new rule should work without debug symbols.

Closes: https://github.com/python-ldap/python-ldap/issues/156

### DIFF
--- a/Misc/python-ldap.supp
+++ b/Misc/python-ldap.supp
@@ -33,7 +33,8 @@
    match-leak-kinds: definite
    fun:malloc
    fun:PL_strdup
-   fun:tlsm_init
+   ...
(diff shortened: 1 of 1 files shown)


encukou:
What's the chance of a false positive here?
Another approach would be to document that the debug symbols are necessary.

tiran:
The changes are pretty low to zero. Also any memory leak in the callstack of ``ldap_set_option`` is a leak inside ``libldap_r``, not a leak in our code.

encukou:
Thanks!

PR #158, opened 2018-01-10
Bump version to 3.0.0b4 and update CHANGES

encukou:


tiran:
How do you feel about calling it release candidate 1 instead of beta 4?

encukou:
Next time. If there is a next time :)

Issue #161, opened 2018-01-10
Test suite is failing on big endian machines


tiran:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

The test suite is failing on big endian machines like PPC64 and S390x. It seems like some code paths are not endian-safe. See https://koji.fedoraproject.org/koji/taskinfo?taskID=24110971

---

Issue description:

```
======================================================================
FAIL: test_pagedresults_decode (Tests.t_ldap_controls_libldap.TestLibldapControls)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-ldap-3.0.0/python2/Tests/t_ldap_controls_libldap.py", line 38, in test_pagedresults_decode
    self.assertEqual(lib.size, SIZE)
AssertionError: 21474836480 != 5
======================================================================
FAIL: test_pagedresults_encode (Tests.t_ldap_controls_libldap.TestLibldapControls)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-ldap-3.0.0/python2/Tests/t_ldap_controls_libldap.py", line 25, in test_pagedresults_encode
    self.assertEqual(pr.encodeControlValue(), lib.encodeControlValue())
AssertionError: '0\x0b\x02\x01\x05\x04\x06cookie' != '0\x0e\x02\x04\x81\xd0\xdf0\x04\x06cookie'
```

Steps to reproduce: 
- run tests on a big endian architecture

Operating system: 
Fedora Linux 27

Python version: 
any

python-ldap version: 
3.0.0-b4

PR #162, opened 2018-01-10
Use correct types for BER en/decode

tiran:
ber_scanf() and ber_printf() "i" format uses ber_int_t. lber_types.h
defines the type as int but Python code assumes the type to be unsigned
long:

```
    #define LBER_INT_T int
    typedef LBER_INT_T ber_int_t;
```

The code was working fine on little endian machines but broke on big
endian machines. ber_int_t is now correctly parsed as signed int.

Fixes: https://github.com/python-ldap/python-ldap/issues/161

### DIFF
--- a/Modules/ldapcontrol.c
+++ b/Modules/ldapcontrol.c
@@ -242,7 +242,7 @@ encode_rfc2696(PyObject *self, PyObject *args)
     BerElement *ber = 0;
     struct berval cookie, *ctrl_val;
     Py_ssize_t cookie_len;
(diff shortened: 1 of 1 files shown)


encukou:
This is not a regression from python-ldap 2.5, right?

If that's the case, it shouldn't hold back the 3.0 release.

tiran:
It's a regression. The bad code has been there for a long time, at least 2011. We didn't notice the bug because we just started to test the feature.

PR #164, opened 2018-01-11
Allow uri and base to be taken from system/user LDAP configuration

sambrightman:
The aim is to reduce the number of parameters that must be passed to
various functions to the minimal set when LDAP is correctly configured
in the environment. In my case, ldap.conf contains at least the URI
and base, and is deployed externally to relevant machines. Duplicating
the configuration in my code would be both inconvenient and more
fragile. I believe this is the case in many common setups.

I think the ideal situation would actually be to re-order some of the
method parameters. People using defaults from the system configuraiton
probably want to only specify e.g. the filter. Putting it first (with
uri/base towards the end), would allow removing the keyword in
calls. However, that is a minor inconvenience versus breaking backward
compatibility.

This is a prototype pull request - in particular:

* I'm not familiar with python-ldap (or really with LDAP in general).
* Docs, Demos, Tests have not been reviewed for potential updates (I
  have the impression that testing could be achieved via environment
  variables and would be willing to add tests).
* I did not know how to treat `urlfetch` and am not 100% that I have caught
  all locations that could be defaulted.


encukou:
A check failed because of a timeout; I restarted it.

I'm not the best person to review the feature, but if no-one else chimes in, I'll look at it next week.

Modules/LDAPObject.c
```diff
@@ -1162,7 +1162,7 @@ l_ldap_search_ext( LDAPObject* self, PyObject* args )
     int msgid;
     int ldaperror;
 
-    if (!PyArg_ParseTuple( args, "sis|OiOOdi",
+    if (!PyArg_ParseTuple( args, "zis|OiOOdi",
```
tiran: Just move the ``|`` to the left and initialize the targets with NULL/0.

Lib/ldap/ldapobject.py
```diff
@@ -748,13 +748,13 @@ def result4(self,msgid=ldap.RES_ANY,all=1,timeout=None,add_ctrls=0,add_intermedi
         resp_data = self._bytesify_results(resp_data, with_ctrls=add_ctrls)
     return resp_type, resp_data, resp_msgid, decoded_resp_ctrls, resp_name, resp_value
 
-  def search_ext(self,base,scope,filterstr=None,attrlist=None,attrsonly=0,serverctrls=None,clientctrls=None,timeout=-1,sizelimit=0):
+  def search_ext(self,base=None,scope=ldap.SCOPE_SUBTREE,filterstr=None,attrlist=None,attrsonly=0,serverctrls=None,clientctrls=None,timeout=-1,sizelimit=0):
```
tiran: Why did you decide to use subtree here?

tiran (changes requested):
Interesting idea. Before this patch can land, you must come up with a test plan, some functional tests and documentation update.

Since it's a new feature, the patch is out-of-scope for 3.0 release.

Lib/ldap/ldapobject.py
sambrightman: It seems the more general (for searches) and is more common in examples.

Modules/LDAPObject.c
tiran: We can make all arguments optional, too. ``z`` is correct to accept None and map it to NULL string.

encukou:
Rebased on current master, in preparation for re-formatting the code.

Modules/LDAPObject.c
mistotebe: This is only called from python-ldap Python code and that will already pass the right arguments in (with a default), so the separator can stay where it is?

Lib/ldap/ldapobject.py
mistotebe: The chief example of an interface with a default for scope is ldapsearch and it picks `-s sub`, so this is being consistent.

tiran (changes requested):
Could you please provide a test case and a ``versionchanged`` entry for your patch?

A test case could work like this:

* create a temporary ldap.conf file
* Run a subprocess with an LDAP query with environment variable ``LDAPCONF``

Modules/options.c
```diff
@@ -163,7 +163,7 @@ LDAP_set_option(LDAPObject *self, int option, PyObject *value)
     case LDAP_OPT_SOCKET_BIND_ADDRESSES:
 #endif
         /* String valued options */
-        if (!PyArg_Parse(value, "s:set_option", &strval))
+        if (!PyArg_Parse(value, "z:set_option", &strval))
```
droideck, line 166: Hmm, when no global BASE is configured, resetting a connection’s OPT_DEFBASE to None raises MemoryError and leaves its previous base unchanged. I think we need to handle this unset-global case explicitly... For example, we can pass the empty DN for this per-connection reset, so restoring the default succeeds. WDYT?

```
previous = conn.get_option(ldap.OPT_DEFBASE)  # None
conn.set_option(ldap.OPT_DEFBASE, "dc=example,dc=com")
conn.set_option(ldap.OPT_DEFBASE, previous)  # Raises MemoryError
```

droideck (changes requested):

Modules/options.c
mistotebe: TBH that sounds like a libldap bug. And it also looks like one... Might have to paper around it?

mistotebe: I've filed [ITS#10606](https://bugs.openldap.org/show_bug.cgi?id=10606) and will put a workaround here, just like you say, if we get null, read the global defbase and if NULL, just pass '' instead.

PR #165, opened 2018-01-11
Make bytes mode TypeError more useful

tiran:
It's a bit hard to understand where the TypeError excactly comes from.
The exception now contains the argument name.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -146,7 +146,7 @@ class SimpleLDAPObject:
         if self.bytes_mode_hardfail:
           raise TypeError(
             "All provided fields *must* be bytes when bytes mode is on; "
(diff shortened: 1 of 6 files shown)


Lib/ldap/ldapobject.py
```diff
@@ -144,7 +144,10 @@ def _bytesify_input(self, arg_name, value):
         return value
       else:
         if self.bytes_mode_hardfail:
-          raise TypeError("All provided fields *must* be bytes when bytes mode is on; got %r" % (value,))
+          raise TypeError(
+            "All provided fields *must* be bytes when bytes mode is on; "
+            "got {}={!r}".format(arg_name, value)
+          )
```
encukou: We did a similar thing with the warnings, but removed the value altogether.
Error messages might end up in logs as well; how would you feel about *only* mentioning the arg_name?

Issue #166, opened 2018-01-12
Warning and error handling in _bytesify_input is too inflexible


tiran:
The way how ``_bytesify_input`` handles errors and warnings is too inflexible:

https://github.com/python-ldap/python-ldap/blob/ad46c1104ea40117f8fb0fd08653517743f94723/Lib/ldap/ldapobject.py#L125-L158

* For ``bytes_mode=None`` (default), the function emits a bytes warning when the input is text, then encodes it as UTF-8 bytes. Warnings are slow and can cause performance regressions (see #133). The warnings are always emitted, even when warnings are disabled and not shown to the user. The warnings module does not allow early opt-out.
* For ``bytes_mode=False`` (strictly future-compatible) there is no warnings mode at all. When the function is called with bytes as input, it always fails with ``TypeError``. TypeErrors make it really, really hard to port an existing application to future compatible API. There should be a mode that emits just warnings.

## Proposal

I suggest that we introduce a ``bytes_warning`` flag:

* ``bytes_warning=BYTES_WARNINGS_SILENT`` emits no warnings and doesn't raise TypeError for ``bytes_mode`` None, True, and False.
* ``bytes_warning=BYTES_WARNINGS_WARN`` emits a warning when input is text for ``bytes_mode=None`` / ``bytes_mode=True`` or input is bytes for ``bytes_mode=False``.
* ``bytes_warning=BYTES_WARNINGS_ERROR`` raises a TypeError when input is text for ``bytes_mode=None`` / ``bytes_mode=True`` or input is bytes for ``bytes_mode=False``.
*  ``bytes_warning=None`` (default) depends on ``bytes_mode``:
    * ``bytes_mode=None`` -> ``BYTES_WARNINGS_WARN``
    * ``bytes_mode=True`` -> ``BYTES_WARNINGS_ERROR``
    * ``bytes_mode=False`` -> ``BYTES_WARNINGS_ERROR``

The flags are defined as:

```
BYTES_WARNINGS_SILENT = 0
BYTES_WARNINGS_WARN = 1
BYTES_WARNINGS_ERROR = 2
```

encukou:
The proposal looks reaonable. I hope to fully get to it next week. (initial nitpick: let's use `bytes_warning='warn'` instead of numbers.)

I'm also thinking about doing this *as* 3.1. I feel the betas aren't useful any more; AFAIK there's only you trying them out.

tiran:
[...]

PR #171, opened 2018-01-22
Add bytes_strictness to allow configuring behavior on bytes/text mismatch

encukou:
See https://github.com/python-ldap/python-ldap/issues/166 for initial discussion, or the changed docs for what is added.

I've used `bytes_strictness` instead of `bytes_warnings`, since warnings is just one of the possible behaviors. The new name isn't perfect either, though.

### DIFF
--- a/Doc/bytes_mode.rst
+++ b/Doc/bytes_mode.rst
@@ -43,37 +43,47 @@ Encoding/decoding to other formats – text, images, etc. – is left to the cal
 The bytes mode
 --------------
 
-The behavior of python-ldap 3.0 in Python 2 is influenced by a ``bytes_mode``
-argument to :func:`ldap.initialize`.
-The argument can take these values:
+In Python 3, text values are represented as ``str``, the Unicode text type.
 
-``bytes_mode=True``: backwards-compatible
+In Python 2, the behavior of python-ldap 3.0 is influenced by a ``bytes_mode``
+argument to :func:`ldap.initialize`:
 
-    Text values returned from python-ldap are always bytes (``str``).
-    Text values supplied to python-ldap may be either bytes or Unicode.
-    The encoding for bytes is always assumed to be UTF-8.
+``bytes_mode=True`` (backwards compatible):
+    Text values are represented as bytes (``str``) encoded using UTF-8.
 
-    Not available in Python 3.
+``bytes_mode=False`` (future compatible):
+    Text values are represented as ``unicode``.
 
-``bytes_mode=False``: strictly future-compatible
-
-    Text values must be represented as ``unicode``.
-    An error is raised if python-ldap receives a text value as bytes (``str``).
-
-Unspecified: relaxed mode with warnings
-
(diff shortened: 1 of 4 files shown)


encukou:
@tiran, whenever you're ready, could you take a look?
(I realize this is in the order of weeks, thanks to a conference)

tiran:
@encukou Next week after DevConf CZ.

Lib/ldap/ldapobject.py
```diff
@@ -107,20 +108,26 @@ def __init__(
...
+                  "Please call initialize(..., bytes_mode=False) explicitly.")
+                bytes_strictness = 'warn'
+        else:
+            if bytes_strictness is None:
+                bytes_strictness = 'error'
+    else:
+        if bytes_mode:
+            raise ValueError("bytes_mode is *not* supported under Python 3.")
+        bytes_mode = False
+        bytes_strictness = 'error'
```
tiran: Maybe raise an exception when ``bytes_strictness`` argument is neither ``'error'`` nor ``None``?

Doc/bytes_mode.rst
```diff
@@ -43,37 +43,47 @@ Encoding/decoding to other formats – text, images, etc. – is left to the cal
...
-    warnings are emitted when they are not.
+While the type of values *returned* from python-ldap is always given by
+``bytes_mode``, the behavior for “wrong-type” values *passed in* can be
+controlled by the ``bytes_strictness`` argument to :func:`ldap.initialize`:
 
-    The warnings are of type :class:`~ldap.LDAPBytesWarning`, which
-    is a subclass of :class:`BytesWarning` designed to be easily
-    :ref:`filtered out <filter-bytes-warning>` if needed.
+``bytes_strictness='error'`` (default if ``bytes_mode`` is specified):
+  A ``TypeError`` is raised.
```
tiran: We should spell out the behavior in form of an example and list affected attributes

> If bytes_mode is True and a text string is encountered, then a TypeError is raised. If bytes_mode is False...

Bytes mode applies to:
* ``dn`` argument
* ``attr`` argument
*  ``newrdn`` argument for ``rename``
* ``who`` and ``cred`` for ``bind``
* ``user``, ``oldpw`` and ``newpw`` for ``passwd``
* ``base``, ``filterstr`` and ``attrlist`` members for ``search*``
* ``mod_type`` (2nd argument) of modlist member for ``add*``, ``modify*``

encukou:
Should document that the strictnedss option only applies to py2; on py3 it's ignored. It's done this way for py2/py3 single-source compatibility.

Doc/bytes_mode.rst
encukou: I linked to text mode from those arguments. Does that look good?

encukou:
@tiran, do you think you'll find some time to review?

encukou:
@tiran, can you get to a review by end of this week?
It's OK if not, or if you need more time. Just please let me know.

PR #173, opened 2018-01-29
Link with lber

tiran:
setup.cfg overrides settings in setup.py. Include lber in list of
libraries in setup.cfg, too.


encukou:
This also needs to be in [the docs](https://python-ldap.org/en/latest/installing.html#example).... But I see it's already there :)

PR #186, opened 2018-03-14
Rewrite cidict and LDAPUrlExtensions in terms of MutableMapping

encukou:
he UserDict API is not specified when it comes to extending:
it is not clear which methods can be safely overridden.

The MutableMapping ABC fixes this problem by providing an explicit
set of abstract methods, in terms of which the rest of the API
is implemented.

### DIFF
--- a/Lib/ldap/cidict.py
+++ b/Lib/ldap/cidict.py
@@ -15,6 +15,7 @@ class cidict(MutableMapping):
     """
     Case-insensitive but case-respecting dictionary.
     """
+    __slots__ = ('_keys', '_data')
 
     def __init__(self, default=None):
         self._keys = {}
--- a/Lib/ldap/cidict.py
+++ b/Lib/ldap/cidict.py
@@ -5,57 +5,58 @@ names of variable case.
 
 See https://www.python-ldap.org/ for details.
 """
+from collections import MutableMapping
 import warnings
 
 from ldap import __version__
 
-from ldap.compat import IterableUserDict
 
+class cidict(MutableMapping):
+    """
+    Case-insensitive but case-respecting dictionary.
+    """
 
-class cidict(IterableUserDict):
-  """
-  Case-insensitive but case-respecting dictionary.
-  """
+    def __init__(self, default=None):
+        self._keys = {}
+        self._data = {}
+        if default:
+            self.update(default)
 
-  def __init__(self,default=None):
-    self._keys = {}
-    IterableUserDict.__init__(self,{})
-    self.update(default or {})
+    # MutableMapping abstract methods
 
-  def __getitem__(self,key):
-    return self.data[key.lower()]
+    def __getitem__(self, key):
+        return self._data[key.lower()]
 
-  def __setitem__(self,key,value):
-    lower_key = key.lower()
-    self._keys[lower_key] = key
-    self.data[lower_key] = value
+    def __setitem__(self, key, value):
+        lower_key = key.lower()
+        self._keys[lower_key] = key
+        self._data[lower_key] = value
 
-  def __delitem__(self,key):
-    lower_key = key.lower()
-    del self._keys[lower_key]
-    del self.data[lower_key]
+    def __delitem__(self, key):
+        lower_key = key.lower()
+        del self._keys[lower_key]
+        del self._data[lower_key]
 
-  def update(self,dict):
-    for key, value in dict.items():
-      self[key] = value
+    def __iter__(self):
+        return iter(self._keys.values())
 
-  def has_key(self,key):
-    return key in self
+    def __len__(self):
+        return len(self._keys)
 
-  def __contains__(self,key):
-    return IterableUserDict.__contains__(self, key.lower())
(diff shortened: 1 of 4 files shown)


Lib/ldap/cidict.py
```diff
@@ -5,96 +5,66 @@
...
+    def __contains__(self, key):
+        return key.lower() in self._keys
+
+    def clear(self):
+        self._keys.clear()
+        self._data.clear()
+
+    # Backwards compatibility
+
+    def has_key(self, key):
```
jdufresne: Just a suggestion, perhaps `.has_key()` should emit a warning and suggest code use the `in` operator instead. After sufficient time has passed, the backwards compatibility shim could be removed.

Lib/ldapurl.py
```diff
@@ -130,58 +132,64 @@ def __ne__(self,other):
...
+            self.update(default)
+
+    def __setitem__(self, name, value):
+        """
+        value
+            Either LDAPUrlExtension instance, (critical,exvalue)
+            or string'ed exvalue
+        """
+        assert isinstance(value, LDAPUrlExtension)
+        assert name == value.extype
```
tiran: While you are fixing the issue, please replace assert with proper checks. asserts are dangerous because they are ignored with ``python -O``.

Lib/ldapurl.py
```diff
@@ -130,58 +132,64 @@ def __ne__(self,other):
...
+            self._data
+        )
 
-  def __eq__(self,other):
-    assert isinstance(other,self.__class__),TypeError(
-      "other has to be instance of %s" % (self.__class__)
-    )
-    return self.data==other.data
+    def __eq__(self,other):
+        assert isinstance(other, self.__class__), TypeError(
```
tiran: Same here, this should return ``NotImplemented`` instead of raising an exception.

Lib/ldapurl.py
```diff
@@ -130,58 +132,64 @@ def __ne__(self,other):
...
+    def __init__(self, default=None):
+        self._data = {}
+        if default is not None:
+            self.update(default)
+
+    def __setitem__(self, name, value):
+        """
+        value
+            Either LDAPUrlExtension instance, (critical,exvalue)
+            or string'ed exvalue
```
tiran: The doc string doesn't match the code, strings are not accepted.

Lib/ldap/cidict.py
```diff
@@ -5,96 +5,66 @@
...
-    temp[elt] = elt
-  for elt in b:
-    temp[elt] = elt
-  return temp.values()
+
+class cidict(MutableMapping):
+    """
+    Case-insensitive but case-respecting dictionary.
+    """
+
```
tiran: Please add ``__slots__ = ('_keys', '_data')`` to avoid an additional ``__dict__`` on every instance.

Lib/ldapurl.py
```diff
@@ -130,58 +132,64 @@ def __ne__(self,other):
...
-  dictionary type
-  """
-
-  def __init__(self,default=None):
-    UserDict.__init__(self)
-    for k,v in (default or {}).items():
-      self[k]=v
-
-  def __setitem__(self,name,value):
+class LDAPUrlExtensions(MutableMapping):
```
tiran: Please add ``__slots__ = ('_data',)`` to avoid an additional ``__dict__`` on every instance.

Lib/ldap/cidict.py
encukou: I still don't think it's worth it. `has_key` is low-maintenance, not a likely source of bugs, and has well-defined semantics. I'd rather not force change for the sake of change.

Lib/ldapurl.py
mistotebe: This looks addressed already but isn't marked as such yet, so the PR is still sitting with "changes requested" on it.

encukou:
@tiran, your comments should be addressed now. Could you re-review?

Lib/ldap/cidict.py
```diff
@@ -5,96 +5,67 @@
...
-def strlist_union(a,b):
-  """
-  Return union of two lists of case-insensitive strings a,b.
-  """
-  temp = cidict()
-  for elt in a:
-    temp[elt] = elt
-  for elt in b:
-    temp[elt] = elt
-  return temp.values()
```
tiran, line 100: The functions are undocumented and untested, but there might still be software using them. How about deprecating them and removing them in a future release?

tiran: https://github.com/python-ldap/python-ldap/pull/336 adds deprecation warning

tiran:
@encukou Could you please rebase your PR?

encukou:
Here it is.

tiran:
LGTM, let's wait for Travis.

encukou:
The Travis job for b1b3b5a91b252afabf9c3d125ef648907dae5527 is green: https://travis-ci.org/github/python-ldap/python-ldap/builds/695164011

Issue #189, opened 2018-03-15
slapd fails to start when SASL support is missing


gabrielbriones:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

Issue description:
I'm working in packaging python ldap for Clear Lunix but slapd tests fails when starting slapd due to not compiled with SASL support.
This is the line which raises the RuntimeException: https://github.com/python-ldap/python-ldap/blob/master/Lib/slapdtest/_slapdtest.py#L407
The exception message looks like this: 

```
*** RuntimeError: ['/usr/bin/ldapwhoami', '-H', 'ldapi://%2Fhome%2Fgbriones%2Fgithub%2Fpython-ldap%2Fpython-ldap-test-56615%2Fldapi', '-Y', 'EXTERNAL', '-Q'] process failed:
b''
b'ldapwhoami: not compiled with SASL support\n'
```

Steps to reproduce: 
1. Build with not SASL support
2. Run tests

Operating system: Clear Linux 21060

Python version: 3.6.4

python-ldap version: 3.0.0

tiran:
I don't have a platform w/o SASL. Please change line 216 to ``self.cli_sasl_external = ldap.SASL_AVAIL`` and run the test suite again.

gabrielbriones:
Thanks tiran, it actually worked.

PR #190, opened 2018-03-15
Disable SASL external when missing SASL support

gabrielbriones:
#189


tiran:
LGTM

Please add ``Closes: https://github.com/python-ldap/python-ldap/issues/189`` to your commit message.

encukou:
Thank you for sharing the fix, @gabrielbriones!

---

> Please add `Closes: https://github.com/python-ldap/python-ldap/issues/189` to your commit message.

I did that when squashing :)

Issue #192, opened 2018-03-23
PasswordPolicyControl.graceAuthNsRemaining falsely returns None


droideck:
While testing 389-ds-base on RHEL 7.5, we've encountered the issue with PasswordPolicyControl.
It doesn't return the remaining number of grace logins (graceAuthNsRemaining).
Openldap util ldapsearch returns the information on the same setup.

Steps to reproduce: 
1. Install 389-ds-base on 389 port and 'dc=example,dc=com' suffix with 'cn=Directory Manager' and 'Secret123' password
2. Run the test case
https://fedorapeople.org/~spichugi/ppolicy_grace_test.py

389-ds-base:
389-ds-base-1.3.7.5-18.el7.x86_64

Operating system: 
Red Hat Enterprise Linux Server release 7.5 (Maipo)

Python: 3.6.3
python-ldap (3.0.0)

Additional information:
The original test case can be found here:
https://pagure.io/389-ds-base/blob/master/f/dirsrvtests/tests/suites/password/pwdPolicy_controls_test.py (test_pwd_expired_grace_limit)

Issue #193, opened 2018-03-23
Don't use getComponentByName on CHOICE type


tiran:
pyasn1's ``getComponentByName()`` method modifies the internal data structure to materialize the component. This changes the value of a CHOICE data structure. The problem caused bug #192.

I talked Ilya about the issue. There are multiple ways to solve the problem:

1. use the native encoder, which turns an ASN.1 data structure in a nice Python data structure, available since pyasn1 0.2.1
2. for choices, replace ``c.getComponentByName("value")`` code with ``"value" in c``, fully supported since pyasn1 0.3.4
3. use ``c.getComponentByName("value", instantiate=False)``, available since pyasn1 0.4.1

I'd prefer to rewrite all pyasn1 handling to native encoder, but that's much more work. Most controls are not covered by proper unit and integration tests, too. Since we already require pyasn1 >= 0.3.7, option 2 is the simplest way to go.

tiran:
More CHOICE:

* https://github.com/python-ldap/python-ldap/blob/master/Lib/ldap/controls/vlv.py#L31-L37 is only use to create a new request, not to decode a control response.
* sync repl is buggy, https://github.com/python-ldap/python-ldap/blob/master/Lib/ldap/syncrepl.py#L317

PR #194, opened 2018-03-23
Fix parsing of PPolicyControl ASN.1 structure

tiran:
Password policy control decoder failed to handle graceAuthNsRemaining
correctly. The warning.getComponentByName('timeBeforeExpiration') call
materialized the time before expiration CHOICE element. The fixed
implementation uses pyasn1's dict interface to check which CHOICE
element is set.

Closes: https://github.com/python-ldap/python-ldap/issues/192
See: https://github.com/python-ldap/python-ldap/issues/193


droideck:
The code looks good to me and it has fixed the issue we had.

PR #195, opened 2018-03-24
Make SlapdObject.root_dn a property

jdufresne:
Allows overriding `SlapdObject.suffix` or `SlapdObject.root_cn` without also requiring `root_dn` be defined. It is now computed from the existing values at runtime. The result is an easier to use `SlapdObject` class.

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -169,7 +169,6 @@ class SlapdObject(object):
     database = 'mdb'
     suffix = 'dc=slapd-test,dc=python-ldap,dc=org'
     root_cn = 'Manager'
-    root_dn = 'cn=%s,%s' % (root_cn, suffix)
     root_pw = 'password'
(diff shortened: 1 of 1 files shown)


jdufresne:
The test failure looks unrelated to this change.

Lib/slapdtest/_slapdtest.py
```diff
@@ -232,6 +231,10 @@ def __init__(self):
         self.clientcert = os.path.join(HERE, 'certs/client.pem')
         self.clientkey = os.path.join(HERE, 'certs/client.key')
 
+    @property
+    def root_dn(self):
+        return 'cn=%s,%s' % (self.root_cn, self.suffix)
```
tiran: Good idea, but please use ``format``:

```
return 'cn={self.root_cn},{self.suffix}'.format(self=self)
```

jdufresne:
Thanks for the review. I've made the suggested change.

PR #196, opened 2018-03-25
In SlapdObject,  build include directives dynamically

jdufresne:
Allows overriding openldap_schema_files without requiring rewriting the conf file.


Lib/slapdtest/_slapdtest.py
```diff
@@ -330,6 +331,12 @@ def gen_config(self):
         }
         return self.slapd_conf_template % config_dict
 
+    def _include_directives(self):
```
tiran: There is no need for an extra function. Instead of a new function, simply create a variable in the ``gen_config`` method and use it as ``'include_directives': include_directives,``.

jdufresne:
Thanks for the review. I've made the suggested change.

PR #197, opened 2018-03-27
Tests: Disable warnings-as-errors for Python 3.4

encukou:
Python 3.4 is in its security fix only phase. We should make sure
the tests pass, but warnings should no longer break CI.


encukou:
This is blocking other PRs, and it's not a code change, so I'm merging without review. Please open an issue or a PR with a fix if you see a problem.

PR #198, opened 2018-03-30
Make SlapdObject a context manager

jdufresne:
Allows for automatic cleanup of resources using the with statement. For example:

```py
   with slapdtest.SlapdObject() as server:
       server.ldapadd(...)
       ...
```

When using `SlapdObject` in a function, it is more convenient and concise than using a `try`/`finally` pattern.

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -164,6 +164,10 @@ class SlapdObject(object):
 
     When a reference to an instance of this class is lost, the slapd
     server is shut down.
+
+    An instance can be used as a context manager. When exiting the context
+    manager, the slapd server is shut down and the temporary data store is
+    removed.
     """
     slapd_conf_template = SLAPD_CONF_TEMPLATE
     database = 'mdb'
@@ -552,6 +556,13 @@ class SlapdObject(object):
             extra_args.append('-r')
         extra_args.append(dn)
         self._cli_popen(self.PATH_LDAPDELETE, extra_args=extra_args)
+
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.stop()
 
 
 class SlapdTestCase(unittest.TestCase):
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -168,6 +168,10 @@ class SlapdObject(object):
     An instance can be used as a context manager. When exiting the context
     manager, the slapd server is shut down and the temporary data store is
     removed.
+
+    .. versionchanged:: 3.1
+
+        Added context manager functionality
     """
     slapd_conf_template = SLAPD_CONF_TEMPLATE
     database = 'mdb'
--- a/Tests/t_slapdobject.py
+++ b/Tests/t_slapdobject.py
@@ -1,0 +1,18 @@
+import unittest
+
+import slapdtest
+
+
+class TestSlapdObject(unittest.TestCase):
+    def test_context_manager(self):
+        with slapdtest.SlapdObject() as server:
+            self.assertIsNotNone(server._proc)
+        self.assertIsNone(server._proc)
+
+    def test_context_manager_after_start(self):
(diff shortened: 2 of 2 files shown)


encukou:
slapdtest is an internal helper for python-ldap tests. I'm worried about adding features that aren't used in the tests.
Specifically, there are no API guarantees: if you use slapdtest elsewhere, we might break your code at any time. (But of course, you're welcome to fix any breakage if the fix makes sense for python-ldap as well.)
Are you OK with that?

jdufresne:
I understand. I'm ok with that, yes.

FWIW, I find slapdtest extremely convenient and useful to tests my application's LDAP interactions. I see a lot of value in it. The module is [documented](https://www.python-ldap.org/en/latest/reference/slapdtest.html), so its use is encouraged to some degree.

Would you like me to add a test? I don't mind adding it and will help make sure the API continues to work.

encukou:
The main thing I'm worried about is feature creep: I'm not sure what the scope of slapdtest should be.
If it's an internal test helper for python-ldap, then the scope is clear – whatever python-ldap needs. I didn't have a problem adding the cert support, for example.
But if it's to be a generic library, we need to somehow define boundaries: what slapdtest is and what it isn't. I'd be fine with defining the scope informally, even as an idea floating in the maintainer's head. But currently, I don't have that :)
For example, should slapdtest be useful without python-ldap? Should it have its own tests? In that case, it would make sense to turn it into a separate project (still under the python-ldap organization).

This particular change is fine; I'll merge it next time I get a block of time for python-ldap.
As long as slapdtest is as an internal helper, it's enough that python-ldap tests pass, no need for an extra test.

tiran:
I share Petr's opinion. For me, the slapdtest package is an internal test helper for python-ldap. We should document the fact, that slapdtest doesn't have the same API stability as the rest of the package.

I see two problems with your PR

1) There is no test case for the context manager protocol. The method are not executed in any test.
2) The behavior of ``__enter__`` is undefined when ``start`` was already called.

jdufresne:
> In that case, it would make sense to turn it into a separate project (still under the python-ldap organization).

If you choose to take this direction. I'm OK with that. Here is an example of another project using `slapdtest` to greatly improve the robustness of its test suite: https://github.com/django-auth-ldap/django-auth-ldap/commit/d72144a654f43b48fc034a915f6b0bd17dc62064

From @encukou 
> no need for an extra test.

From @tiran 
> There is no test case for the context manager protocol. The method are not executed in any test.

I'm still happy to write a test if it is desired, but I'm getting a mixed message on whether or not it is desired.

> The behavior of `__enter__` is undefined when start was already called.

It behaves exactly the same as calling `start()` twice. Meaning, the second call will not start a second slapd server due to the [`if _proc is None` guard](https://github.com/python-ldap/python-ldap/blob/23905322fdb7084f81e487162aa2b09e6b702787/Lib/slapdtest/_slapdtest.py#L430). When the context manager exits, the server is always stopped.

jdufresne:
I have updated the PR to include:

- A test for the context manager
- A test for the context manager after calling `.start()`
- Added documentation about using the context manager

encukou:
> > In that case, it would make sense to turn it into a separate project (still under the python-ldap organization).
>
> If you choose to take this direction. I'm OK with that. Here is an example of another project using `slapdtest` to greatly improve the robustness of its test suite: django-auth-ldap/django-auth-ldap@d72144a

OK. Let's go in that direction. https://github.com/python-ldap/python-ldap/issues/203

> I'm still happy to write a test if it is desired, but I'm getting a mixed message on whether or not it is desired.

Always better to have a test! It's not absolutely required if we think of sldaptest as only an internal testing tool.

encukou:
In code I'd write today, anything that could have `start` and `stop` methods would be a context manager *first*. With explicit start/stop calls it's too easy to forget the clean-up.
Of course, unittest, coming from an age when we didn't have context managers, is built around explicit setup/teardown calls. That's fine, no need to rebuild the world when something better/safer/easier comes along, but it's no reason to not add support for the new thing.

I want to merge this. @tiran, do want to comment?

PR #200, opened 2018-03-30
Move import statements to top level

jdufresne:
For import statements that can safely exist at the top level of the file, move them there.

In `Lib/slapdtest/_slapdtest.py`, having imports local to functions unnecessarily resulted in duplicates.

This style is also more in line with PEP8:

https://www.python.org/dev/peps/pep-0008/#imports

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

### DIFF
--- a/Lib/ldap/asyncsearch.py
+++ b/Lib/ldap/asyncsearch.py
@@ -8,6 +8,8 @@ import ldap
 
 from ldap import __version__
 
+import ldif
+
 SEARCH_RESULT_TYPES = {
   ldap.RES_SEARCH_ENTRY,
   ldap.RES_SEARCH_RESULT,
@@ -269,7 +271,6 @@ class LDIFWriter(FileWriter):
   """
 
   def __init__(self,l,writer_obj,headerStr='',footerStr=''):
-    import ldif
     if isinstance(writer_obj,ldif.LDIFWriter):
       self._ldif_writer = writer_obj
     else:
--- a/Lib/ldap/schema/subentry.py
+++ b/Lib/ldap/schema/subentry.py
@@ -4,9 +4,16 @@ ldap.schema.subentry -  subschema subentry handling
 See https://www.python-ldap.org/ for details.
 """
 
+import copy
+
 import ldap.cidict,ldap.schema
 
+from ldap.compat import urlopen
 from ldap.schema.models import *
+
+import ldapurl
+import ldif
+
 
 SCHEMA_CLASS_MAPPING = ldap.cidict.cidict()
 SCHEMA_ATTR_MAPPING = {}
@@ -256,7 +263,6 @@ class SubSchema:
     Get a schema element by name or OID with all class attributes
     set including inherited class attributes
     """
-    import copy
     inherited = inherited or []
     se = copy.copy(self.sed[se_class].get(self.getoid(se_class,nameoroid)))
     if se and hasattr(se,'sup'):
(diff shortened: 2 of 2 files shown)


Lib/slapdtest/_slapdtest.py
```diff
@@ -279,7 +279,6 @@ def _cleanup_rundir(self):
         """
         # cleanup_rundir() is called in atexit handler. Until Python 3.4,
         # the rest of the world is already destroyed.
-        import os, os.path
```
tiran: Don't remove the local import here. The requirement is documented in the comment above the import.

jdufresne: Thanks for the review.

I'm curious, do you have any documentation, resource, or experiment that demonstrates why this import and comment is necessary?

I did experimentation with Fedora's Python 2.7.14. To experiment, I applied the following diff to force all cleanup during the atexit handler:

```patch
diff --git a/Lib/slapdtest/_slapdtest.py b/Lib/slapdtest/_slapdtest.py
index bb03a1f..f38bf6e 100644
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -441,6 +441,7 @@ class SlapdObject(object):
         """
         Stops the slapd server, and waits for it to terminate and cleans up
         """
+        print('stop')
         if self._proc is not None:
             self._log.debug('stopping slapd with pid %d', self._proc.pid)
             self._proc.terminate()
@@ -575,6 +576,6 @@ class SlapdTestCase(unittest.TestCase):
         cls.server = cls.server_class()
         cls.server.start()
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.server.stop()
+    # @classmethod
+    # def tearDownClass(cls):
+    #     cls.server.stop()
```

The atexit handler never had a problem or failure using the module level `os` import. From what I can tell, this comment seems unnecessary or outdated. 

How can I demonstrate to myself this import is necessary?

Lib/slapdtest/_slapdtest.py
```diff
@@ -277,9 +277,6 @@ def _cleanup_rundir(self):
         """
         Recursively delete whole directory specified by `path'
         """
-        # cleanup_rundir() is called in atexit handler. Until Python 3.4,
```
tiran: The code is there for a specific reason. The ``cleanup_rundir`` is designed to be executed in an ``atexit`` handler. By the time the handler runs, the rest of the world may already be gone. You must keep both imports as long as python-ldap supports Python < 3.4.

jdufresne: I understand the motivation for why this was originally added.

Having understood it, I'm asking for a way to demonstrate its usefulness to myself. [In my testing](https://github.com/python-ldap/python-ldap/pull/200#discussion_r178464223), whether or not the atexit handler used the top level import or the local import never made a practical difference. Neither failed, raised an exception, nor produced a warning.  If this is really necessary or useful, it should be possible to demonstrate it as such. How?

Right now, my concern is that it was added to avoid a issue that doesn't actually occur in practice.

tiran: It occurs in practice and I ran into the issue pretty often during testing.

The problem not easy to reproduce. The bug occurs only on Python 2 and requires a combination of uncollectable reference cycles and bad luck. The problem with atexit handlers is a well-known and understood issue. If you search the internet for "Python atexit NoneType has no attribute", you'll find a lot of mentions.

To avoid the issue, a Python 2 atexit handlers must be self-contained and are not allowed to depend on any global variables or state. This especially includes module state, because modules are cleared first.

jdufresne:
Thanks. I've reverted the changes to the atexit handler.

PR #202, opened 2018-04-02
Doc: Replace Mac OS X -> macOS, this is now the official name

intgr:


Doc/faq.rst
```diff
@@ -112,9 +112,9 @@ telling Lib/ldap.py and Lib/ldap/schema.py are not found::
     The messages above are falsely produced by DistUtils.
     Don't worry about it.
 
-.. _install-macosx:
+.. _install-macos:
```
encukou: The slug `install-macosx` is an arbitrary string; any similarity to names like “Mac OS X” is just a mnemonic :)

People on the Internet might already be linking to [it](https://python-ldap.readthedocs.io/en/latest/faq.html#install-macosx), so please don't change it if you can avoid it.

intgr: Right, I wasn't entirely sure about that myself. I changed it back now.

Issue #85, opened 2017-12-04
Global get/set_option() functions do not release the GIL


tiran:
The global and local get/set option functions share the same code. The code uses ``LDAP_BEGIN_ALLOW_THREADS`` but that macro only releases the GIL for local settings. It does nothing when LDAP object is NULL for global settings.

https://github.com/python-ldap/python-ldap/blob/c84b4faf429431cb4d4ef42a156b62d1105f1758/Modules/LDAPObject.h#L34-L48

PR #206, opened 2018-04-05
Release GIL around global get/set option call

tiran:
ldap.set_option() and ldap.get_option() now properly release the GIL
around potentially blocking calls.

Also fixes a problem with make indent by replacing
Py_BEGIN/END_ALLOW_THREADS with macros that are terminated by a
semicolon.

Fixes: https://github.com/python-ldap/python-ldap/issues/85

### DIFF
--- a/Modules/functions.c
+++ b/Modules/functions.c
@@ -15,13 +15,18 @@ l_ldap_initialize(PyObject *unused, PyObject *args)
     char *uri;
     LDAP *ld = NULL;
     int ret;
+    PyThreadState *save;
 
     if (!PyArg_ParseTuple(args, "s:initialize", &uri))
         return NULL;
 
-    Py_BEGIN_ALLOW_THREADS ret = ldap_initialize(&ld, uri);
-    Py_END_ALLOW_THREADS if (ret != LDAP_SUCCESS)
+    save = PyEval_SaveThread();
+    ret = ldap_initialize(&ld, uri);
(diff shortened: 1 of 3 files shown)


Modules/LDAPObject.h
```diff
@@ -47,4 +47,13 @@ extern LDAPObject *newLDAPObject(LDAP *);
...
 
+#define BEGIN_ALLOW_THREADS                                         \
+    do {                                                            \
+        PyThreadState *_save;                                       \
+        _save = PyEval_SaveThread()                                 \
+
+#define END_ALLOW_THREADS                                           \
+        PyEval_RestoreThread(_save);                                \
+    } while(0)
+
```
encukou: This isn't something I'd want to maintain in python-ldap.

I think you can follow Python's `Py_*_ALLOW_THREADS` macros with a semicolon. Or is there some reason you can't?
(If not, I'd rather propose that in Python, and possibly keep a backport here for current/past Python versions.)

tiran:
I pushed a new version of the patch. A semicolon after Py_BEGIN_ALLOW_THREADS may work with some compilers, but break with others. I have now spelled out the function calls. python-ldap requires threading any way.

encukou:
> A semicolon after `Py_BEGIN_ALLOW_THREADS` may work with some compilers, but break with others.

Is that worth fixing in CPython, or do you think it would be unacceptable there?

PR #209, opened 2018-04-12
More detailed documantation on some pages

somay:
I'm porting a Python 3 application from some other ldap client library to python-ldap, which requires some investigation of the internal implementation (not only Python code but also C).
Here is the outcome of my such code reading.

### DIFF
--- a/Doc/reference/ldap-modlist.rst
+++ b/Doc/reference/ldap-modlist.rst
@@ -20,12 +20,11 @@ The :mod:`ldap.modlist` module defines the following functions:
    in the result at all.
 
    The value parts of the *entry* dictionary must be any of a list of
-   :py:class:`bytes` objects, a :py:class:`bytes` object, or :py:const:`None` (in Python 3).
+   :py:class:`bytes` objects or a :py:class:`bytes` object (in Python 3).
    The :py:class:`bytes` object is treated as a list which contains
    single :py:class:`bytes` object.
    These :py:class:`bytes` data are passed to the C interface without
    any conversion.
-   :py:const:`None` means :c:data:`NULL` in C (no values).
 
 
 
--- a/Doc/reference/ldap-modlist.rst
+++ b/Doc/reference/ldap-modlist.rst
@@ -19,6 +19,15 @@ The :mod:`ldap.modlist` module defines the following functions:
    names which shall be ignored completely. Attributes of these types will not appear
    in the result at all.
 
+   The value parts of the *entry* dictionary must be any of a list of
+   :py:class:`bytes` objects, a :py:class:`bytes` object, or :py:const:`None` (in Python 3).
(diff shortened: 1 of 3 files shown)


Doc/reference/ldap-modlist.rst
```diff
@@ -19,6 +19,15 @@ The :mod:`ldap.modlist` module defines the following functions:
    names which shall be ignored completely. Attributes of these types will not appear
    in the result at all.
 
+   The value parts of the *entry* dictionary must be any of a list of
+   :py:class:`bytes` objects, a :py:class:`bytes` object, or :py:const:`None` (in Python 3).
+   The :py:class:`bytes` object is treated as a list which contains
+   single :py:class:`bytes` object.
+   These :py:class:`bytes` data are passed to the C interface without
+   any conversion.
+   :py:const:`None` means :c:data:`NULL` in C (no values).
```
encukou: That's not correct; addModlist doesn't accept `None` as a value.
Could you show how you tested this?

somay: Oh, sorry. I looked only `Tuple_to_LDAPMod()` in Modules/LDAPObject.c .

You might be able to eliminate the code below.
https://github.com/python-ldap/python-ldap/blob/master/Modules/LDAPObject.c#L142-L144

somay: I wrote a test on `None` case.
Please tell me if you don't want to include it in your test suites.

Doc/reference/ldap-modlist.rst
```diff
@@ -19,6 +19,14 @@ The :mod:`ldap.modlist` module defines the following functions:
    names which shall be ignored completely. Attributes of these types will not appear
    in the result at all.
 
+   The value parts of the *entry* dictionary must be any of a list of
+   :py:class:`bytes` objects or a :py:class:`bytes` object (in Python 3).
```
somay: I start to think that accepting non-list `bytes` objects is possibly a bug.
The code below seems to assume values of entry dicts (the `value` variable in the code) are `list` (or `tuple`), not `bytes`.
Resultant `attrvaluelist` is always a list.
`attrvaluelist` is however never used thereafter and the original `value` variable is used instead.

https://github.com/python-ldap/python-ldap/blob/master/Lib/ldap/modlist.py#L20-L23

encukou: To me, the intended behavior for single `bytes` (and for `None items) it's also a bit of a mystery. Let's be careful here, to not break users' code.

somay:
Sorry for being inactive for a long time.
I currently don't have enough time to write tests to validate my documentation.
I'll open more fine-grained PRs with enough tests when I'm really ready.
I appreciate your reviews anyway.
Thank you for maintaining this project.

PR #213, opened 2018-04-28
Call ._bytesify_input() on value arg in  .compare_ext()

### LINKED ISSUE #212: UnicodeEncodeError: SimpleLDAPObject.compare_ext doesn't call _bytesify_input on value arg
Issue description:

In `SimpleLDAPObject.compare_ext()`, with `bytes_mode=False`, the arguments `dn` and `attr` are converted via `._bytesify_input()`. The argument `value`, is not, but I believe is should be. Code is here:

https://github.com/python-ldap/python-ldap/blob/f23f7901f5343581fcfbc3a5db8a745d8adb4e3f/Lib/ldap/ldapobject.py#L513-L515

Without the conversion, I receive errors that look like:

```
...
  File "/home/jon/devel/python-ldap/Lib/ldap/ldapobject.py", line 534, in compare_s
    return self.compare_ext_s(dn,attr,value,None,None)
  File "/home/jon/devel/python-ldap/Lib/ldap/ldapobject.py", line 519, in compare_ext_s
    msgid = self.compare_ext(dn,attr,value,serverctrls,clientctrls)
  File "/home/jon/devel/python-ldap/Lib/ldap/ldapobject.py", line 516, in compare_ext
    return self._ldap_call(self._l.compare_ext,dn,attr,value,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
  File "/home/jon/devel/python-ldap/Lib/ldap/ldapobject.py", line 313, in _ldap_call
    result = func(*args,**kwargs)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xdf' in position 7: ordinal not in range(128)
```

Steps to reproduce: 

1. Use Python 2.7
2. Initialize python-ldap with `bytes_mode=False`
3. Call `.compare_s()` where all 3 arguments contain non-ascii characters

Operating system: 

Fedora release 27

Python version: 

Python 2.7

python-ldap version: 

master

jdufresne:
Fixes #212

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -513,6 +513,7 @@ class SimpleLDAPObject:
     if PY2:
         dn = self._bytesify_input('dn', dn)
         attr = self._bytesify_input('attr', attr)
+        value = self._bytesify_input('value', value)
     return self._ldap_call(self._l.compare_ext,dn,attr,value,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
 
   def compare_ext_s(self,dn,attr,value,serverctrls=None,clientctrls=None):
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -71,6 +71,10 @@ dn: cn=Foo4,ou=Container,%(suffix)s
 objectClass: organizationalRole
 cn: Foo4
 
+dn: cn=Ƒøø5,ou=Container,%(suffix)s
+objectClass: organizationalRole
+cn: Ƒøø5
+
 """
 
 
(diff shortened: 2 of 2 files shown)


encukou:
Is this actually correct? LDAP values should always be bytes in python-ldap. Encoding is left to the caller.
AFAIK `_bytesify_input` should only be called on things that LDAP defines to be UTF-8 encoded text.

tiran:
That's a good point. Attribute values are defined as bytes.

jdufresne:
Thanks for the review. 

Does that mean the tests in f3ff4a320a21a3ac42bc7587eaed09c4f9b2f9f5 are in fact incorrect, they should instead be:

```py
result = l.compare_s('cn=Foo1,%s' % base, 'cn', b'Foo1')
```

(`b` prefix on value argument)

jdufresne:
Per feedback above, closing. This was my misunderstanding.

PR #214, opened 2018-04-28
Drop u prefix from literals in files that import unicode_literals

jdufresne:
For files that have `from __future__ import unicode_literals`, the u prefix is unnecessary. String literals without a prefix are already Unicode literals. By dropping the prefix, uses a modern Python3 syntax style.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -260,7 +260,7 @@ class SimpleLDAPObject:
     else:
       # It's a list of referals
       # Example value:
-      # [u'ldap://DomainDnsZones.xxxx.root.local/DC=DomainDnsZones,DC=xxxx,DC=root,DC=local']
+      # ['ldap://DomainDnsZones.xxxx.root.local/DC=DomainDnsZones,DC=xxxx,DC=root,DC=local']
       return [self._maybe_rebytesify_text(referal) for referal in result_value]
 
   def _bytesify_results(self, results, with_ctrls=False):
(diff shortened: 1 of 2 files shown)


tiran:
I'd rather drop ``from __future__ import unicode_literals`` and make ``u''`` / ``b''`` explicit.

encukou:
The problem is that if `unicode_literals` is not in *all* files, you need to check the top of the file to see what an unprefixed string means. The `u` prefix is unambiguous.

> If it makes the change better, I can add `unicode_literals` to all files so this is universally true across the project.

If you'd be willing to do that, it would definitely make the change better!

tiran:
I'd still prefer to remove ``unicode_literals`` and use an explicit ``u`` prefix for strings. Several Python core developers including Nick and me agree, that unicode literal mode was a mistake, see http://python-future.org/unicode_literals.html

encukou:
I agree that `unicode-literals` is not ideal, especially because it offers no way to write a “native string” literal. Also, at this point we're *not* trying to make py2/py3 behavior more similar. Please avoid behavior changes in Python 2 – people should spend time porting away from it, not adapting to new changes.

> IMO, the project should be looking forward to future Python -- with compatibility workarounds for old Pythons -- and not the other way around of preferring older styles with workarounds for modern Pythons.

I currently have no intention of supporting Python 2 past 2020. Perhaps I should make that clear somewhere.

---

However, `unicode-literals` everywhere *would* be an improvement over the current state (assuming it would introduce no functional changes), so, whoever does the work gets to decide :)
(Please coordinate to avoid doing the work twice, though.)

jdufresne:
I don't mind closing.

PR #215, opened 2018-04-28
Use builtin unittest methods assertIsNone & assertIsNotNone

jdufresne:
These methods were introduced in Python 2.7 and 3.1. Available in all supported environments. Can remove the local reimplementation.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNone
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsNone


encukou:
Thank you, and sorry for the delay in review!

PR #216, opened 2018-04-28
Fix assertTrue(result, _ldap.RES_BIND) to assertEqual()

jdufresne:
Fix a pattern where `assertTrue(result, _ldap.RES_BIND)` was mistakenly used in place of `self.assertEqual(result, _ldap.RES_BIND)`.

Mistakes introduced in commit 162cedf71ba62b3944d18a27ed206d55e679f966.


encukou:
Thank you, and sorry for the delay in review!

Issue #218, opened 2018-05-24
Segfault with multi-entry attributes


tiran:
LDAP supports multi-valued entries. In almost all common cases ``ldap_first_attribute`` / ``ldap_next_attribute`` return an attribute once. In very rare cases like FreeIPA's ``ipaAllowedToPerform;write_keys``, an entry can appear more than once in the result. python-ldap accumulates the multi-entry attribute in a single list.

However the implementation has a reference counting bug. It uses ``PyDict_GetItem`` and treats the object as owned object. However ``PyDict_GetItem`` only returns a borrowed reference. Application code must either ``Py_INCREF`` the reference to own it or never ``Py_DECREF`` the result.

The bug was never noticed before, because our tests don't trigger this particular behavior with OpenLDAP server. Although I tried, I wasn't able to reproduce the issue with OpenLDAP. But 389-DS can return an attribute multiple times, which causes python-ldap to segfault. Under Python 2, it doesn't segfault but the list is garbage.

mikeporterinmd:
Using openldap, I was able to get two copies of the same attribute, with different value lists.  Setting through using ldap_get_values_len didn't work correctly.  It tended to find the first set of values.  So, I was getting:

udService: 2fac-cas
udService: 2fa-required
udService: 2fac-cas
udService: 2fa-required

instead of:

udService: 2fac-cas
udService: 2fa-required
udService: mail.udel.edu

I'm not exactly sure how to fix this.  However, I can say that the openldap ldapsearch command did return the correct results.  I'll check its source code.  The "fix" was to just more or less recreate the entry and it worked.  The issue can be created by adding an entry to an existing attribute.

mikeporterinmd:
Hmmm... since ldap_get_values only accepts an attribute name and not a handle, I don't see how this could ever work.  Not sure of a solution except to make sure that the LDAP server does not return multiples of the same attribute.  All values should be on one instance of the attribute.  Not sure if it makes sense to remove the code that checks the output dict or not.

PR #219, opened 2018-05-24
Fix ref counting bug in LDAPmessage_to_python

tiran:
PyDict_GetItem() returns a borrowed reference. The code later
Py_DECREF() the reference, which causes a segfault at a later point in
time.

Fixes: https://github.com/python-ldap/python-ldap/issues/218


encukou:
Oh, I read https://github.com/python-ldap/python-ldap/issues/218 after this one; there you say this isn't reproducible on OpenLDAP.

encukou:
Would you mind adding a code comment saying why this case is not covered by tests?
e.g. just linking to the issue

PR #222, opened 2018-06-10
In docs, use intersphinx to link to Python documentation

jdufresne:
The intersphinx extension can generate automatic links to the documentation of objects in other projects. For complete details on intersphinx, see:

http://www.sphinx-doc.org/en/master/ext/intersphinx.html

### DIFF
--- a/Doc/bytes_mode.rst
+++ b/Doc/bytes_mode.rst
@@ -93,12 +93,11 @@ to be given as well.
 Porting recommendations
 -----------------------
 
-Since end of life of Python 2 is coming in a few years,
-projects are strongly urged to make their code compatible with Python 3.
-General instructions for this are provided `in Python documentation`_ and in
-the `Conservative porting guide`_.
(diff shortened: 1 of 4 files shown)


tiran:
The intersphinx plugin and intersphinx links require a network connection to build documentation. In the past it was no longer possible to build documentation without a working internet connection. Is this still the case?

encukou:
AFAIK, distro packages are compiled without an internet connection, so that all sources (even of the docs) can be audited. The machine that builds the docs is not connected, but the resulting docs are then used on normal users' machines.

Technically, this is a problem for distros and/or intersphinx maintainers – ideally the intersphinx indexes should be made available offline, and could even point to offline documentation if the distro makes python-ldap docs depend on Python docs.

That said, the links this replaces are pretty stable; I see no rush getting this in.
Can we reach out on the mailing list, and preferably Fedora/Debian maintainers explicitly, to see what they have to say here?

PR #225, opened 2018-06-19
filter out None from attribute values at modlist.addModlist()

somay:
Obviously a bug (unused variable and inconsistent behaviour with `modifyModlist()`).
Fix with a test.

### DIFF
--- a/Lib/ldap/modlist.py
+++ b/Lib/ldap/modlist.py
@@ -20,7 +20,7 @@ def addModlist(entry,ignore_attr_types=None):
     # Eliminate empty attr value strings in list
     attrvaluelist = [item for item in value if item is not None]
     if attrvaluelist:
-      modlist.append((attrtype, value))
+      modlist.append((attrtype, attrvaluelist))
   return modlist # addModlist()
 
 
--- a/Lib/ldap/modlist.py
+++ b/Lib/ldap/modlist.py
@@ -9,7 +9,7 @@ from ldap import __version__
 import ldap
 
 
-def addModlist(entry,ignore_attr_types=None):
+def addModlist(entry, ignore_attr_types=None):
   """Build modify list for call of method LDAPObject.add()"""
   ignore_attr_types = {v.lower() for v in ignore_attr_types or []}
   modlist = []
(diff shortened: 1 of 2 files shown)


encukou:
(Sorry for the delay; I was on vacation)

I'm tempted to agree. However, I'm not sure what the correct behavior around `None` is, and I'd like that clarified and documented when changing the code.

It seems that the code was indended to handle only `[None]` specially, but not something like `[None, a_value]`. I'd rather warn/error for such cases, rather than silently skip them.
Do you know if/why `[None, a_value]` should be treated as `[a_value]` in modlists? (`None` corresponds to a `NULL` berval in C, but I know nothing about NULL bervals – I'm not much of a LDAP expert.)
@tiran, maybe you can give some background here?

somay:
Everyone needs vacations :+1: .

You need not worry about `None` case because values such as `[b'test', None]` are already rejected by `Tuple_to_LDAPMod()` in `LDAPobject.c` when calling `LDAPObject.add()` and so on.
It's not the matter of the LDAP specs.
See the code snippet to reproduce and the traceback below.


```python
import ldap
import ldap.modlist


def main():
    ldap_object = ldap.initialize("ldap://need.not.connect")

    entry = {'uid': [b'test', None]}

    add_modlist = ldap.modlist.addModlist(entry)
    ldap_object.add("uid=test,dc=example,dc=com", add_modlist)


if __name__ == '__main__':
    main()
```

```
  File "add.py", line 14, in <module>
    main()
  File "add.py", line 10, in main
    ldap_object.add_ext_s("uid=test,dc=example,dc=com", add_modlist)
  File "/home/somay/code/python-ldap/Lib/ldap/ldapobject.py", line 413, in add_ext_s
    msgid = self.add_ext(dn,modlist,serverctrls,clientctrls)
  File "/home/somay/code/python-ldap/Lib/ldap/ldapobject.py", line 410, in add_ext
    return self._ldap_call(self._l.add_ext,dn,modlist,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
  File "/home/somay/code/python-ldap/Lib/ldap/ldapobject.py", line 313, in _ldap_call
    result = func(*args,**kwargs)
TypeError: ('Tuple_to_LDAPMod(): expected a byte string in the list', None)
```

This `if` branch raises it.

https://github.com/python-ldap/python-ldap/blob/master/Modules/LDAPObject.c#L174-L178

So filtering out `None` is sensible unless you are concerned about other cases such as directly manipulating the results of `addModlist()`.
I don't think we need to keep backward-compatibility with such hacky use cases.

I fixed `bytes` case without deprecation warnings. I think adding it is not my job.

encukou:
> So filtering out None is sensible unless you are concerned about other cases such as directly manipulating the results of addModlist().

Is it? You shouldn't have `None` in the modlist, but to me that means we should raise an error when it's found, rather than silently ignore it.

> I don't think we need to keep backward-compatibility with such hacky use cases.

Do we not?
If we want people to use newer versions of python-ldap, it should work with their old applications.

somay:
Okay. I got it.
It would be nicer to me if the error is raised by `addModlist()`, than by `add_ext()` and also by `modifyModlist()`.
But we must take compatibility into account.
I keep in mind it next time I contribute.

encukou:
One option to preserve backcompat but discourage questionaple input is to raise a warning.

But yes, python-ldap is an old project, used in old systems, so we need to be careful.

PR #227, opened 2018-07-05
Docs: Fix internal set|get_option func|meth references

jdufresne:
Previous syntax did not create links and appeared as raw rst syntax.

https://www.python-ldap.org/en/python-ldap-3.0.0/reference/ldap.html#options


encukou:
Good catch; thank you!

Issue #228, opened 2018-07-09
Integer value for ENOTCONN hardcoded in tests


iv-m:
### Issue description:

I'm trying to build python-ldap on mipsel. The build runs fine, but the tests that check for `errno 107` fail. Here is a relevant piece of the test logs:

```
======================================================================
FAIL: test_errno107 (Tests.t_cext.TestLdapCExtension)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/RPM/BUILD/python-module-ldap-3.0.0/Tests/t_cext.py", line 748, in test_errno107
    self.fail("expected errno=107, got %d" % errno)
AssertionError: expected errno=107, got 134

======================================================================
FAIL: test004_errno107 (Tests.t_ldapobject.Test00_SimpleLDAPObject)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/RPM/BUILD/python-module-ldap-3.0.0/Tests/t_ldapobject.py", line 462, in test004_errno107
    self.fail("expected errno=107, got %d" % errno)
AssertionError: expected errno=107, got 134

======================================================================
FAIL: test004_errno107 (Tests.t_ldapobject.Test01_ReconnectLDAPObject)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/RPM/BUILD/python-module-ldap-3.0.0/Tests/t_ldapobject.py", line 462, in test004_errno107
    self.fail("expected errno=107, got %d" % errno)
AssertionError: expected errno=107, got 134

----------------------------------------------------------------------
```

The reason is that on this system no possible `errno` value equals to 107, and `ENOTCONN` is defined as 134:

```
$ python -c 'import errno; print errno.ENOTCONN'
134
```

Of course, this command prints `107` on my x86_64 laptop.

As far as I understand, the integer value of `ENOTCONN` is not standard and indeed can vary from platform to platform.

### Steps to reproduce: 
```
git clone https://github.com/python-ldap/python-ldap
cd python-ldap
tox -e py27
```
Operating system:  mipsel port of ALT Linux Sisyphus.

Python version: 2.7, 3.5, 3.6

python-ldap version: both 3.1.0 and current git master have this issue.

PR #229, opened 2018-07-09
tests: Use system-specific ENOTCONN value

iv-m:
The integer values for `errno` are  not standard and can vary from
platform to platform, so the tests should rely on the constants
provided in `errno` module.

Closes: #228


encukou:
Looks good, thank you!

PR #231, opened 2018-07-16
Feature request: Less strict LDIF parser mode

xi:
The thunderbird addressbook export sometimes results in broken ldif files. That is why I would like to add a less strict parser mode that ignores recoverable parse errors.

I understand that it would be better to fix this thunderbird. Still, I think it is a useful feature, e.g. for debugging.

In the non-strict mode, the parser in my implementation just silently ignores the errors. It would be better to have some kind of logging. I did not add this yet as I was not sure how logging works in other parts of this library.

### DIFF
--- a/Lib/ldif.py
+++ b/Lib/ldif.py
@@ -254,7 +254,8 @@ class LDIFParser:
     ignored_attr_types=None,
     max_entries=0,
     process_url_schemes=None,
-    line_sep='\n'
+    line_sep='\n',
+    strict=True
   ):
     """
     Parameters:
@@ -271,6 +272,8 @@ class LDIFParser:
         is ignored completely.
     line_sep
         String used as line separator
+    strict
+        If set to ``False``, recoverable parse errors will be ignored.
     """
     self._input_file = input_file
     # Detect whether the file is open in text or bytes mode.
@@ -279,6 +282,7 @@ class LDIFParser:
     self._process_url_schemes = list_dict([s.lower() for s in (process_url_schemes or [])])
     self._ignored_attr_types = list_dict([a.lower() for a in (ignored_attr_types or [])])
     self._last_line_sep = line_sep
+    self._strict = strict
     self.version = None
     # Initialize counters
     self.line_counter = 0
@@ -292,6 +296,10 @@ class LDIFParser:
       self._last_line = self._readline()
     except EOFError:
       self._last_line = ''
+
+  def _recoverable_error(self, msg):
+    if self._strict:
+      raise ValueError(msg)
 
   def handle(self,dn,entry):
     """
@@ -422,12 +430,12 @@ class LDIFParser:
           (not self._max_entries or self.records_read<self._max_entries):
       # Consume first line which must start with "dn: "
       if k!='dn':
-        raise ValueError('Line %d: First line of record does not start with "dn:": %s' % (self.line_counter,repr(k)))
+        self._recoverable_error('Line %d: First line of record does not start with "dn:": %s' % (self.line_counter,repr(k)))
       # Value of a 'dn' field *has* to be valid UTF-8
       # k is text, v is bytes.
       v = v.decode('utf-8')
       if not is_dn(v):
-        raise ValueError('Line %d: Not a valid string-representation for dn: %s.' % (self.line_counter,repr(v)))
+        self._recoverable_error('Line %d: Not a valid string-representation for dn: %s.' % (self.line_counter,repr(v)))
       dn = v
       entry = {}
       # Consume second line of record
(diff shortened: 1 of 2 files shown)


encukou:
I'm not a fan of this feature, since it basically brings undefined behavior to python-ldap.
What counts as a "recoverable" error? What data do you get from such invalid output – for example, can future versions of python-ldap produce different output?
I'm also not convinced that this feature is particularly useful. For debugging, couldn't you fix the LDIF file instead?

xi:
I understand the concern. I still think this is a useful feature.

> What counts as a "recoverable" error? What data do you get from such invalid output – for example, can future versions of python-ldap produce different output?

In my implementation I treat four cases as "recoverable":

1. First line is always treated as dn, validating that the key is actually "dn" is optional.
2. Validating dn value is optional.
3. Validating changetype is optional.
4. Validating mod-op is optional.

I think that the behavior should be stable and maybe even documented.

In cases 2-4, the validation in non-strict-mode would be up to the calling code. In my case, Thunderbird generates LDIF files with invalid dn values. As I do not use the dn, I could simply ignore that. I expect there are more usecases like mine.

Maybe my approach is not the best one. I think my main issue with the current behavior is that the parser will crash if anything in the input is invalid. It is not even possible to skip invalid entries. I am not saying that invalid input should be silently accepted — there should definitely be warnings. But a bit of "be liberal in what you accept" could make the LDIFParser more useful.

encukou:
Is there a bug open for Thunderbird? LDIF is a well-specified format; if the DNs aren't valid, it shouldn't be called LDIF export.

xi:
Sure, there are plenty (e.g. [154554](https://bugzilla.mozilla.org/show_bug.cgi?id=154554), [168241](https://bugzilla.mozilla.org/show_bug.cgi?id=168241), [230334](https://bugzilla.mozilla.org/show_bug.cgi?id=230334)). But that is not the point. There will be other tools that output invalid LDIF.

encukou:
Sorry, I remain unconvinced.
There are many ways LDIF can be invalid. The four cases in this implementation might not be enough. Or they might be too much – are there really tools that output invalid changetypes or mod-ops?

Another thing is that some of the checks are asserting things that the code later relies on. For example, if the modop string is invalid, the `modop = MOD_OP_INTEGER[k]` would be skipped. What should the numeric code then *be* in this case?

Lib/ldif.py
```diff
@@ -528,7 +536,7 @@ def parse_change_records(self):
             try:
               modop = MOD_OP_INTEGER[k]
```
encukou: (I'm referring to this line)

xi:
So are you opposed to the idea in general, or just to this specific implementation? If it is the former, I think it's fine just to close this pull request. However, if it is the letter, I would be interested in some pointers of how the implementation should be changed.

encukou:
> So are you opposed to the idea in general, or just to this specific implementation? 

I think it's not a good feature *for python-ldap*. This project focuses on implementing standards; the RFE is too experimental.
But it looks like it's not a bad idea *per se*, which is why I'm trying to understand the use cases. I'd like to suggest some better place for the feature, instead of just saying "don't do that here".

If it's just Thunderbird exports, it would be better to fix Thunderbird. If that's impossible now, maybe create a script specifically to clean-up Thunderbird exports?

xi:
In my opinion the current implementation mixes two things: Parsing and validation. It is currently not possible to only do one or the other.

For me this feels like an HTML parser then crashes if there is a typo in the text. It is important to check for typos, but that should be a separate step. I feel like for you, all validation that is currently done is a fundamental part of parsing LDIF. I can completely understand that perspective as it is what the spec defines. But it is not my perspective.

>  I'd like to suggest some better place for the feature, instead of just saying "don't do that here".

The trouble is that this library does not provide any hooks for this kind of feature. Implementing it somewhere else would mean to fork the whole thing. I have actually done that in https://github.com/xi/ldif3 a while ago. But I would like to deprecate that fork.

encukou:
> The trouble is that this library does not provide any hooks for this kind of feature. Implementing it somewhere else would mean to fork the whole thing.

I hear you. Certainly, this PR would make it easier to write a cleaner for Thunderbird.

However, is the goal to turn the ldif module into a generic base for parsers of LDIF-like formats? Such a thing is too easy to get wrong (e.g. to make it only useful for one use case only). And `python-ldap` can't afford to make mistakes here – once an API is in place, it's very hard to change it.
If not, the scope needs to be definer better: "add special cases to make it possible to handle Thunderbird exports" is not very convincing; "make debugging easier" is too vague.

> In my opinion the current implementation mixes two things: Parsing and validation. It is currently not possible to only do one or the other.

Yes. The implementation also mixes in a step of *interpreting* the data when it converts modop strings to integer codes.

Would it be possible to actually do these steps separately? Say, one function to do line-folding and base64-encoding, others to do the interpreting, validation and packaging into dicts?
Such a chain would make more sense to me than a system where a part can be turned off. (To test such a system, you should essentially run the testsuite twice; once with `strict=True` and then with `False`. As more such features are added, that would become intractable – and if we're going for "generic base for LDIF-like formats" rather than "workaround for Thunderbird", then more features probably *will* be needed.) 

> For me this feels like an HTML parser then crashes if there is a typo in the text. It is important to check for typos, but that should be a separate step. 

Well, you could also compare it to an image reader rejecting an image with a size of `120` by `banana` pixels...

> I feel like for you, all validation that is currently done is a fundamental part of parsing LDIF. I can completely understand that perspective as it is what the spec defines. But it is not my perspective.

Yes, this is where we think differently.
Since python-ldap's parsing and validation are mixed together, I can't commit to maintain the ability to switch off part of the code – at least not without clear scope, widely applicable use cases, thorough tests & documentation.
If your use case is just Thunderbird exports, maintaining a fork will probably be less work than aligning this with python-ldap's parser. In such a fork, you *can* afford to make design mistakes that might prevent generalizing this beyond Thunderbird.

So, I'll close this PR now. We can still discuss it or reopen, of course.

xi:
Thanks for the in-depth reply, I really appreciate that.

Maybe I can come up with a cleaner solution to my problem and open a new pull request. Otherwise I will keep using my fork.

Issue #226, opened 2018-06-22
AttributeError: 'module' object has no attribute '_trace_level'


misery:
Issue description:
After an upgrade from 3.0.0 to 3.1.0 I cannot login into my review board as the ldap module throws an error. After an downgrade everything works fine.

Steps to reproduce: 
```
[...]
  File "/opt/reviewboard/dist/lib/python2.7/site-packages/reviewboard/accounts/backends.py", line 1008, in authenticate
    for uri, connection in self.get_ldap_connections(userdomain):

  File "/opt/reviewboard/dist/lib/python2.7/site-packages/reviewboard/accounts/backends.py", line 935, in get_ldap_connections
    connection = ldap.initialize(ldap_uri)

  File "/opt/reviewboard/dist/lib/python2.7/site-packages/ldap/functions.py", line 85, in initialize
    return LDAPObject(uri,trace_level,trace_file,trace_stack_limit,bytes_mode)

  File "/opt/reviewboard/dist/lib/python2.7/site-packages/ldap/ldapobject.py", line 99, in __init__
    self._trace_level = trace_level or ldap._trace_level

AttributeError: 'module' object has no attribute '_trace_level'
```

Operating system: Linux
Python version: 2.7.14
python-ldap version: 3.1.0 and 3.0.0

Xophmeister:
I am able to replicate this -- and came to post an issue -- but only when running Python with optimisations enabled (i.e., `python -O` or `python -OO`).

Steps to reproduce:

In `test.py`:
```python
import ldap
ldap.initialize("ldap://my.ldap.host")
```

Then at the command line:
```
$ python -O test.py
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    server = ldap.initialize("ldap://my.ldap.host")
  File "/Users/ch12/Projects/play/python/ldap/.venv/lib/python3.6/site-packages/ldap/functions.py", line 85, in initialize
    return LDAPObject(uri,trace_level,trace_file,trace_stack_limit,bytes_mode)
  File "/Users/ch12/Projects/play/python/ldap/.venv/lib/python3.6/site-packages/ldap/ldapobject.py", line 99, in __init__
    self._trace_level = trace_level or ldap._trace_level
AttributeError: module 'ldap' has no attribute '_trace_level'
```

This is with Python 3.6.5 on macOS 10.11.6, with python-ldap 3.1.0

encukou:
[...]

PR #234, opened 2018-07-30
Lib/ldap: Provide _trace_level and _trace_file in non-debug mode

encukou:
All use of these attributes *should* be guarded by `if __debug__`.
However, that's not always the case. Providing different API based
on `__debug__` is unnecessarily fragile.

This is intended as a quick fix for a maintenance release (3.1.1). More coming up in another PR.

Fixes: https://github.com/python-ldap/python-ldap/issues/226


ww3pl:
I'd also love to see this reviewed.

PR #235, opened 2018-07-31
Add Python 3.8-dev to Tox config

encukou:
Let's catch issues early – both in python-ldap and CPython.

With all relevant compilers supporting C99, Python 3.6+ makes use of [some C99 features](https://www.python.org/dev/peps/pep-0007/#c-dialect), and in 3.8-dev `//` comments found their way to headers included from `Python.h`.
I'd like to keep checking C90 compat for older Python versions, hence a bit of indirection to set the -std flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,13 @@ matrix:
         - WITH_GCOV=1
       dist: xenial
       sudo: true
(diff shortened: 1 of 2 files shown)


encukou:
There are no code changes, so I'll merge this.

PR #247, opened 2018-10-08
Add support for X-ORIGIN in schema element class ObjectClass

droideck:
389 Directory Server uses X-ORIGIN ObjectClass attribute for determining if the ObjectClass is 'user defined' or not. 389 Directory Server UI currently depends on it.
Also, X-ORIGIN is already supported for AttributeType schema element. So it makes sense to make it consistent across these two schema elements.

### DIFF
--- a/Lib/ldap/schema/models.py
+++ b/Lib/ldap/schema/models.py
@@ -126,7 +126,7 @@ class ObjectClass(SchemaElement):
   sup
     This list of strings contains NAMEs or OIDs of object classes
     this object class is derived from
-  x-origin
+  x_origin
     This string contains the X-ORIGIN text which is typically used to indicate
     the source of the associated schema element. It can a list of strings
   """
@@ -230,7 +230,7 @@ class AttributeType(SchemaElement):
   sup
     This list of strings contains NAMEs or OIDs of attribute types
     this attribute type is derived from
-  x-origin
+  x_origin
     This string contains the X-ORIGIN text which is typically used to indicate
     the source of the associated schema element. It can a list of strings
   """
--- a/Lib/ldap/schema/models.py
+++ b/Lib/ldap/schema/models.py
@@ -128,7 +128,7 @@ class ObjectClass(SchemaElement):
     this object class is derived from
   x-origin
     This string contains the X-ORIGIN text which is typically used to indicate
-    the source of the associated schema element
+    the source of the associated schema element. It can a list of strings
   """
   schema_attribute = u'objectClasses'
   token_defaults = {
@@ -141,7 +141,7 @@ class ObjectClass(SchemaElement):
     'ABSTRACT':None,
     'MUST':(()),
     'MAY':(),
-    'X-ORIGIN':(None,)
+    'X-ORIGIN':()
   }
 
   def _set_attrs(self,l,d):
@@ -150,7 +150,7 @@ class ObjectClass(SchemaElement):
     self.desc = d['DESC'][0]
     self.must = d['MUST']
     self.may = d['MAY']
-    self.x_origin = d['X-ORIGIN'][0]
+    self.x_origin = d['X-ORIGIN']
     # Default is STRUCTURAL, see RFC2552 or draft-ietf-ldapbis-syntaxes
     self.kind = 0
     if d['ABSTRACT']!=None:
@@ -173,7 +173,7 @@ class ObjectClass(SchemaElement):
     result.append({0:' STRUCTURAL',1:' ABSTRACT',2:' AUXILIARY'}[self.kind])
     result.append(self.key_list('MUST',self.must,sep=' $ '))
     result.append(self.key_list('MAY',self.may,sep=' $ '))
-    result.append(self.key_attr('X-ORIGIN',self.x_origin,quoted=1))
+    result.append(self.key_list('X-ORIGIN',self.x_origin,quoted=1))
     return '( %s )' % ''.join(result)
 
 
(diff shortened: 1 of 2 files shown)


encukou:
I checked how the attribute looks on the object, and all looks fine. I suggest adding this extra test. Could you quickly review it?

droideck:
AttributeType also doesn't have info about 'X-ORIGIN' in its docstring. Should I update it too?

Test looks good to me.

encukou:
On Thu, Oct 25, 2018, 18:05 Simon Pichugin <notifications@github.com> wrote:

> AttributeType also doesn't have info about 'X-ORIGIN' in its docstring.
> Should I update it too?
>
That would be great.

droideck:
I've added the docstrings.

One more issue popped up. It is described here -
https://pagure.io/389-ds-base/pull-request/50005#comment-67932

So should I implement it? Are you okay with it?
And should I do it in this PR?

encukou:
As I see it, the problem with X-ATTR is that it's not really standardized anywhere. Is that right?
Does any other server use it as a multi-valued attribute, or is this 389-DS specific behavior?

droideck:
> As I see it, the problem with X-ATTR is that it's not really standardized anywhere. Is that right?
Yes, it's right.

> Does any other server use it as a multi-valued attribute, or is this 389-DS specific behavior?
At least, Oracle's Directory Server uses it the same way we are. :)

Anyway, if we'll add the multi-value support it wouldn't hurt the existing functionality.
The same way we have NAME now.
It will be NAME ( 'nsAIMid' 'nscpaimscreenname' ) for multi-values and NAME 'cosspecifier' for single-valued.

Also, currently, if we have X-ORIGIN ( 'my objectclass' 'user defined' ), the existing python-ldap implementation will just print the first value, which is wrong.

droideck:
I've added the functionality I mentioned before and a test for it (also, I used FreeIPA LDIFs which have multi-valued X-ORIGIN, btw).

Please, check.

droideck:
Are there any objections? :)
Or we can merge? The tests pass.

encukou:
Tests pass, but there's standard to verify that the behavior is correct. I can't find anything to suggest that there's agreement between LDAP servers on what X-ORIGIN actually means.
Is there such a standard or some other neutral documentation that could be referenced here?

vashirov:
> LDAP does not define any standard schema extensions, but many directory servers accept any properly-formatted extension as a means of annotating the schema element. For example, one of the most commonly-used extension types is “X-ORIGIN”, which is typically used to indicate the source of the associated schema element, and the string “X-ORIGIN ‘RFC 4519’” might be used to indicate that the associated schema element is defined in RFC 4519.

as seen on 
[1] https://ldap.com/schema-element-extensions/
[2] https://ldapwiki.com/wiki/LDAP%20Schema%20Element%20Extensions

Another usage example in LDAP server implementation:
https://github.com/ForgeRock/opendj-community-edition/blob/master/resource/schema/00-core.ldif

Lib/ldap/schema/models.py
```diff
@@ -126,6 +126,9 @@ class ObjectClass(SchemaElement):
   sup
     This list of strings contains NAMEs or OIDs of object classes
     this object class is derived from
+  x-origin
+    This string contains the X-ORIGIN text which is typically used to indicate
+    the source of the associated schema element. It can a list of strings
```
encukou:
```suggestion
    the source of the associated schema element. It should be be a tuple of strings.
```

Lib/ldap/schema/models.py
```diff
@@ -126,6 +126,9 @@ class ObjectClass(SchemaElement):
   sup
     This list of strings contains NAMEs or OIDs of object classes
     this object class is derived from
+  x-origin
```
encukou:
```suggestion
  x_origin
```

Lib/ldap/schema/models.py
droideck: Agree. But it has a small typo - "be be"

droideck: Also, should we change other docstrings then for the consistency? For example, 'may' also should be a tuple but the docstring states 'This list of strings':

    may
        This list of strings contains NAMEs or OIDs of additional attributes
        an entry of the object class may have

droideck:
> I'm missing a test that setting x_origin works, and a test of the behavior of setting x_origin (to values like None, list, tuple, or single string).

Sure.
I've searched for any examples in other tests but it looks to me that we don't test 'set' operation of any particular attributes (like self.names = d['NAME'], self.desc = d['DESC'][0], self.must = d['MUST'], etc.)

So if I missed something and it's there (or you have tools for that) feel free point me to it :)
Thanks

Lib/ldap/schema/models.py
encukou: I guess "[sequence](https://docs.python.org/3/glossary.html#term-sequence)" is the exact thing we want here.
Perhaps write it out as "sequence (list/tuple)", to aid people who don't actively know that definition.

encukou:
Well, new features need tests, even if old ones don't have them.
Right now I don't know how they would be written, but I can look into it tomorrow (and, in the process, actually write the tests).

Lib/ldap/schema/models.py
droideck: Actually, list wouldn't work because the SchemaElement class explicitly requires a 'tuple'.

``` python
class SchemaElement:
    def key_list(self,key,values,sep=' ',quoted=0):
        assert type(values)==tuple,TypeError("values has to be a tuple, was %r" % values)
```

droideck:
> Well, new features need tests, even if old ones don't have them.
> Right now I don't know how they would be written, but I can look into it tomorrow (and, in the process, actually write the tests).

Sure, I agree. Thank you very much!

encukou:
Can I ask for a quick check of these?

droideck:
Looks great! Thank you!

Issue #249, opened 2018-10-17
ldap.initialize() unknown keyword parameter bytes_strictness


jdennis:
ldap.initialize() is documented as having a bytes_strictness parameter as seen here: http://www.python-ldap.org/en/latest/reference/ldap.html#ldap.initialize.

It needs to be added to the parameter list here: https://github.com/python-ldap/python-ldap/blob/ec41070d3fa1d00cb272f86e0c8ca55db51bdf62/Lib/ldap/functions.py#L68

also the docstring, and finally in the call to LDAPObject().

The LDAPObject (SimpleLDAPObject in ldapobject.py) already supports the bytes_strictness parameter.

As an initial work-around one can do this:
```python
conn = ldap.initialize(url, bytes_mode=False)
conn.bytes_strictness = 'silent'
```
bytes_strictness defaults to 'error' thus defeating the automatic type conversion described here: http://www.python-ldap.org/en/latest/bytes_mode.html#bytes-mode

PR #250, opened 2018-10-25
 Make initialize() pass extra keyword arguments to LDAPObject 

encukou:
Fixes: #249 

Also, documentation for these should now match the code.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -76,7 +76,9 @@ class NO_UNIQUE_ENTRY(ldap.NO_SUCH_OBJECT):
 
 class SimpleLDAPObject:
   """
-  Drop-in wrapper class around _ldap.LDAPObject
+    This basic class wraps all methods of the underlying C API object.
+
+    The arguments are same as for the :func:`~ldap.initialize()` function.
   """
 
   CLASSATTR_OPTION_MAPPING = {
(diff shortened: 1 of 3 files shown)


encukou:
@tiran, could you find some time for a review?

Lib/ldap/ldapobject.py
```diff
@@ -76,7 +76,9 @@ class NO_UNIQUE_ENTRY(ldap.NO_SUCH_OBJECT):
 
 class SimpleLDAPObject:
   """
-  Drop-in wrapper class around _ldap.LDAPObject
+    This basic class wraps all methods of the underlying C API object.
```
tiran: Please also indent the leading and trailing ``"""``.

Lib/ldap/ldapobject.py
```diff
@@ -1057,15 +1059,20 @@ def get_naming_contexts(self):
...
-  In case of server failure (ldap.SERVER_DOWN) the implementations
-  of all synchronous operation methods (search_s() etc.) are doing
-  an automatic reconnect and rebind and will retry the very same
-  operation.
-
-  This is very handy for broken LDAP server implementations
-  (e.g. in Lotus Domino) which drop connections very often making
-  it impossible to have a long-lasting control flow in the
-  application.
+    :py:class:`SimpleLDAPObject` subclass whose synchronous request methods
```
tiran: Please also indent the leading and trailing ``"""``.

Lib/ldap/ldapobject.py
encukou: That would mean re-indenting the entire class, which is too much for this PR. So I un-indented the contents instead.

Issue #208, opened 2018-04-09
Response controls are not attached to exceptions


somay:
Issue description:

According to [RFC 4512 Section 4.1.1](https://tools.ietf.org/html/rfc4511#section-4.1.1), response controls can be attatched also to unsuccessful response messages.
But exceptions `LDAPObject.result()` or such raise don't have such information.
Are these any way to get response controls even if operations are not succeeded?

I'm encountering this problem when trying to interpret password policy response control from OpenLDAP 2.4 server.
Per [draft-behera-ldap-password-policy Section 8.1.2.3.2](https://tools.ietf.org/html/draft-behera-ldap-password-policy-10#section-8.1.2.3.2), error controls such as `passwordExpired` are conveyed with return codes indicating operation errors.

Steps to reproduce: 

1. enable password policy overlay on OpenLDAP 2.4 server
1. set a non-zero value to `pwdMaxAge` and `userPassword` to `pwdAttribute` of password policy ldap entry
1. create an user (entry) with `userPassword` attribute
1. wait for expiring
1. call `LDAPObject.simple_bind_s()` on the user (which raises `ldap.INVALID_CREDENTIALS`)

Operating system: Linux

Python version: 3.5.2

python-ldap version: 3.0.0

tiran:
Is the error raises by ``simple_bind`` or ``simple_bind_s``? Please post the track back.

somay:
Traceback:

```
Traceback (most recent call last):
  File "poc.py", line 77, in <module>
    dn = create_wait_and_bind(TREE, lo)
  File "poc.py", line 47, in create_wait_and_bind
    result = lo.simple_bind_s(dn, new_password, serverctrls=request_controls)
  File "/home/somay/pytmp/venv/lib/python3.5/site-packages/ldap/ldapobject.py", line 444, in simple_bind_s
    resp_type, resp_data, resp_msgid, resp_ctrls = self.result3(msgid,all=1,timeout=self.timeout)
  File "/home/somay/pytmp/venv/lib/python3.5/site-packages/ldap/ldapobject.py", line 749, in result3
    resp_ctrl_classes=resp_ctrl_classes
  File "/home/somay/pytmp/venv/lib/python3.5/site-packages/ldap/ldapobject.py", line 756, in result4
    ldap_result = self._ldap_call(self._l.result4,msgid,all,timeout,add_ctrls,add_intermediates,add_extop)
[...]

Issue #177, opened 2018-02-23
Exceptions don't include the message id they are an exception for


gwvandesteeg:
**Issue description:**
When using the async methods and polling for results using the result(msg=RES_ANY), an exception can get raised since one of the async calls generated caused the error.
The exception raised does not include the message id of the request that caused the exception.
Having 10 outstanding requests and receiving an exception for one (ie. ldap.ALREADY_EXISTS) and not knowing which request this was for since we're not receiving the msgid in the response.

**Steps to reproduce:** 
connect to python
alternate an add and a search query using the same query three times (each) using the async methods (or randomise the order)
poll for results until you've received a response for each outstanding request.
all the searches should complete, one of the adds should complete, and two of the adds should've failed, but which add.. which message ids.

**Operating system:** 
Linux

**Python version:** 
2.7+

**python-ldap version:** 
all

encukou:
Thank you for reporting the issue!
Would you be willing to a propose fix, or perhaps a (failing) test case to illustrate how you would like this to be solved?

tiran:
This should be easy to solve by changing ``PyObject* LDAPerror(LDAP*, char* msg);`` to ``PyObject* LDAPerror(LDAP*, char* msg, int msgid);``. We also need a value to signal that we don't have a value msgid. A msgid is a positive int except for two special cases: ``LDAP_RES_ANY`` is defined as ``-1`` and ``LDAP_RES_UNSOLICITED`` as ``0``. Any value small than -1 should do the trick.

gwvandesteeg:
That is basically the same idea I had when looking at the code base,  initialise msgid to None and then set it if we get one,  or a negative value will do just as well for no message id present.

spaceone:
I would like to see this as well :-)

PR #251, opened 2018-11-16
Attach response controls to exceptions

somay:
Fixes: #208 

`exception.args[0]['ctrls']` is a list of the form of `(oid: str, criticality: int, berval: bytes)`.

I've choose `ctrls` instead of `controls` as the key of the dictionary because the other keys are also abbreviated like `desc`, `info`.

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -654,12 +654,3 @@ class PPolicyEnabledSlapdObject(SlapdObject):
                 'ppolicy_use_lockout'])
         },
     )
-
-
-class PPolicyEnabledSlapdTestCase(SlapdTestCase):
-    """
-    A subclass of :py:class:`SlapdTestCase`, which uses
-    :py:class:`PPolicyEnabledSlapdObject` as the slapd controller.
-    """
-
-    server_class = PPolicyEnabledSlapdObject
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -623,34 +623,3 @@ class SlapdTestCase(unittest.TestCase):
     @classmethod
     def tearDownClass(cls):
         cls.server.stop()
-
-
-class PPolicyEnabledSlapdObject(SlapdObject):
-    """
-    A subclass of :py:class:`SlapdObject` with password policy enabled.
-    Note that this class has no actual password policy configuration entries.
-    It is the job of the users of this class to define
-    the default password policies on their own.
-    The dn of the default is :attr:`.default_ppolicy_dn` of this class.
-    """
-
-    openldap_schema_files = (
-        'core.schema', 'ppolicy.schema'
-    )
-    modules = (
-        'ppolicy',
-    )
-
-    default_ppolicy_dn = "cn=default-ppolicy,%(suffix)s" % {
-        'suffix': SlapdObject.suffix
-    }
-
-    overlays = (
-        {
-            'name': 'ppolicy',
-            'configuration': "\n".join([
-                'ppolicy_default "{}"'.format(default_ppolicy_dn),
-                # let slapd tell the clients that they are locked out
-                'ppolicy_use_lockout'])
-        },
-    )
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -612,17 +612,10 @@ class SlapdTestCase(unittest.TestCase):
         """
         return a LDAPObject instance after simple bind
         """
-        ldap_conn = self._make_ldap_object(**kwargs)
-        ldap_conn.simple_bind_s(who or self.server.root_dn, cred or self.server.root_pw)
-        return ldap_conn
-
-    def _make_ldap_object(self, **kwargs):
-        """
-        return an unbound LDAPObject instance with common ldap options.
-        """
         ldap_conn = self.ldap_object_class(self.server.ldap_uri, **kwargs)
         ldap_conn.protocol_version = 3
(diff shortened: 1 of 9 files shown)


Lib/slapdtest/_slapdtest.py
```diff
@@ -596,3 +623,43 @@ def setUpClass(cls):
...
+    )
+
+
+class PPolicyEnabledSlapdTestCase(SlapdTestCase):
+    """
+    A subclass of :py:class:`SlapdTestCase`, which uses
+    :py:class:`PPolicyEnabledSlapdObject` as the slapd controller.
+    """
+
+    server_class = PPolicyEnabledSlapdObject
```
encukou: Is this necessary? `PPolicyEnabledSlapdTestCase` needs to be subclassed; the subclass can set its `server_class` directly.

Lib/slapdtest/_slapdtest.py
```diff
@@ -596,3 +623,43 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         cls.server.stop()
+
+
+class PPolicyEnabledSlapdObject(SlapdObject):
```
encukou: Do you think this class will be needed for more than just one test? If not, would it be better to define the class just in the file for that test?

Lib/slapdtest/_slapdtest.py
```diff
@@ -187,6 +191,9 @@ class SlapdObject(object):
         'core.schema',
     )
 
+    modules = ()
+    overlays = ()
```
encukou: Since these attributes can be overridden in subclasses, they should be documented.

Lib/slapdtest/_slapdtest.py
```diff
@@ -582,10 +602,17 @@ def _open_ldap_conn(self, who=None, cred=None, **kwargs):
         """
         return a LDAPObject instance after simple bind
         """
+        ldap_conn = self._make_ldap_object(**kwargs)
+        ldap_conn.simple_bind_s(who or self.server.root_dn, cred or self.server.root_pw)
+        return ldap_conn
+
+    def _make_ldap_object(self, **kwargs):
```
encukou: If subclasses are supposed to call this, it should be documented.
Perhaps it would be better to add an overridable `bind=True` class attribute?

Modules/constants.c
```diff
@@ -104,6 +104,11 @@ LDAPerror(LDAP *l, char *msg)
             ldap_memfree(matched);
         }
 
+        if (pyctrls != NULL) {
+            PyDict_SetItemString(info, "ctrls", pyctrls);
+            /* Py_XDECREF(pyctrls) must be called on caller side */
```
encukou: Needing to call `Py_XDECREF(pyctrls)` is implied by Python's calling convention – it's a borrowed reference. No need to point it out here.

Tests/t_ldapobject.py
```diff
@@ -75,6 +76,72 @@
...
+            ldap_conn.simple_bind_s(user_dn, wrong_password)
+
+        with self.assertRaises(ldap.INVALID_CREDENTIALS) as cm:
+            ldap_conn.simple_bind_s(
+                user_dn, password,
+                serverctrls=[ldap.controls.ppolicy.PasswordPolicyControl()])
+
+        controls = cm.exception.args[0]['ctrls']
+        pp = ldap.controls.DecodeControlTuples(controls)[0]
+        self.assertEqual(pp.error, 1)  # error == 1 means AccountLockout
```
encukou: Should we have a constant for `AccountLockout` somewhere?

encukou: Could you also check `len(ldap.controls.DecodeControlTuples(controls))`?

Tests/t_ldapobject.py
```diff
@@ -75,6 +76,72 @@
...
+cn: {cn}
+userPassword: {password}'''.format(cn=cn, dn=user_dn, password=password)
+        )
+
+        ldap_conn = self._make_ldap_object(bytes_mode=False)
+
+        # Firstly cause a bind failure to lock out the account
+        with self.assertRaises(ldap.INVALID_CREDENTIALS):
+            wrong_password = 'wrong' + password
+            ldap_conn.simple_bind_s(user_dn, wrong_password)
```
encukou: Could you also check here that `ctrls` is empty?

somay: checked.
Thank you for quick and detailed review.

somay:
I haven't finished yet.
I'll comment explicitly when I'm ready for your review.

mistotebe:
I have started related work on this and #177 in https://github.com/mistotebe/python-ldap/tree/exceptions, @somay could you have a look and see if these two could be reconciled (I'm happy to do that)?

tiran:
Petr and I are at PyCon until end of next week. It may take two weeks until we are able to look into your PR.

I noticed some unrelated improvements like a memory leak fix and some changes to slapdtest. Could you please file separate bugs and PRs for these? Thanks

mistotebe:
@tiran sure, hope you're having fun, either of you attending this year's LDAPCon?

The linked branch is a work in progress and I've been rewriting it along the way, I guess I can split things at some point when they've settled.

Also, do you think you'll have some time to moderate the email I sent to the mailing list?

somay:
@mistotebe sorry for long inactivity.
I'm also glad if you handle this issue in addition to your one.
Actually, I have little time to finish this PR (sorry) because I will quit my current LDAP related job soon.
Feel free to reuse my code modification and documentation.
I will try to review your draft in this week, although the scope of your PR seems much wider than my scope.

encukou:
Thanks to @mistotebe's work in https://github.com/python-ldap/python-ldap/pull/288, response controls will be available under `ctrls`.

Issue #255, opened 2018-12-21
NameError: name 'basestring' is not defined


cavazquez:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

Issue description:

when I create SSSRequestControl(ordering_rules=['-uidNumber']  its fail because "basestring" in __init__ method not exists in python3.

tips: set basestring = str

Steps to reproduce: 

Operating system: 

Python version: 3.7

python-ldap version: 3.1.0

PR #259, opened 2019-01-28
Python 3 has no basestring, use str instead

mistotebe:
Fixes: #255


mistotebe:
Yes, the test looks ok.

PR #260, opened 2019-01-30
Add Doc/requirements.txt for Read the Docs build

encukou:
The documentation build uses docstrings from the code, and so it needs the code to be importable.
We already use a fake _ldap module for cases where the C compiler is not available, like on Read the Docs.

Another issue on Read the Docs is that the `pyasn1` dependency needs to be installed. This can't be done automatically via setup.py install, as that would attempt to build C code.

Instead, add a documentation-only "requirements.txt", which we can point Read the Docs to.

### DIFF
--- a/Doc/requirements.txt
+++ b/Doc/requirements.txt
@@ -1,0 +1,2 @@
+pyasn1
+pyasn1_modules


encukou:
I made a private docs build to verify this solves the problem.
This is not a code change and it's needed to get documentation building.
I am using my admin powers to merge without review.

Issue #256, opened 2019-01-08
Return ldap.initialize differ of the documentation


eamanu:
Hello everybody!

Reading the documentation https://www.python-ldap.org/en/latest/reference/ldap.html#ldap.LDAPObject this tell that and ldap.LDAPObject is returned by ldap.initialize, but currently `ldap.ldapobject.SimpleLDAPObject` is returned. 

I think that this must be clarify on doc. 

Regards!

Operating system: Linux eamanu 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux 

Python version: Python 2.7.15rc1

python-ldap version: 3.1.0

tiran:
``ldap.ldapobject.LDAPObject`` is an alias for ``ldap.ldapobject.SimpleLDAPObject``. The documentation incorrectly refers to both classes as ``ldap.LDAPObject`` but the ldap top level name space does not provide these attributes.

eamanu:
> `ldap.ldapobject.LDAPObject` is an alias for `ldap.ldapobject.SimpleLDAPObject`. The documentation incorrectly refers to both classes as `ldap.LDAPObject` but the ldap top level name space does not provide these attributes.

That's right.  My unittest failed because I run ` self.assertIsInstance(ldap.initialize(url), ldap.LDAPObject)`. 

I think that in documentation need to say the real class' name

Issue #205, opened 2018-04-05
What is the correct way to change LDAPObject from SimpleLDAPObject?


somay:
python-ldap version: 3.0.0
Python version: 3.5.2

The doc says `ldap.initialize()` uses `ldap.ldapobject.LDAPObject` and also says

> Internally LDAPObject is set to SimpleLDAPObject by default.

It seems to me that *by default* implies that `LDAPObject` can be changes to `ReconnectLDAPObject` or custom class inheriting `ldap.resiter.ResultProcessor` or any other valid classes and then I call `ldap.initialize()` and successfully get a object of my registered class.
The way to change the *default* is not documented though.

I found that assigning a desired class to `ldap.function.LDAPObject` causes the aforementioned behaviour.
But this method seems tricky to me and I wonder it is your expected usage of this library.

It would be great if the things below is clarified in the doc.
Being obvious also through API design would be superb.

* What is the recommended way to use classes other than `SimpleLDAPObject`?
* Which is recommended whether calling `ldap.initialize()` or directly instantiating `ldap.ldapobject.SimpleLDAPObject` or such?

tiran:
``ldap.initialize`` just creates an instance of ``LDAPObject``, which is an alias for ``SimpleLDAPObject``. It's basically a one-liner and doesn't do any extra work. I recommend that you use ``ldap.initialize`` if you want default behavior. In case you need a reconnecting LDAP client or Resiter, then instantiate the classes directly.

@encukou we should clarify this in the documentation.

By the way, I'm going to obsolete Resiter and include the feature into ``LDAPObject``, see #204.

somay:
Thank you a lot @tiran for answering very quickly!
So eventually the issue you mentioned will be resolved and I needn't use my custom class. Great improvement!

somay:
I here list the causes of my confusion for you to choose direction of documetation improvement.

My confusion resulted from:

* `ldap.initialize()` is on very top of documentation.
* `ldap.initialize()` is more easy to import than `ldap.ldapobject`.
* Where `*LDAPObject` is from (actually from `ldap.ldapobject`) is not (as far as I read) documented. 
* `ldap.ldapobject` is not in table of contents in left menu.

encukou:
Re-opening this, since we do need to change the documentation

PR #261, opened 2019-01-30
Doc: Clarify and improve documentation of ldap.ldapobject.LDAPObject

encukou:
### Use the full module in ldap.ldapobject.LDAPObject Doc

Previously, the class showed up as `ldap.LDAPObject`, which is wrong.

<s>Note that the documentation on Read the Docs is currently not building (see https://github.com/python-ldap/python-ldap/pull/260). The `SimpleLDAPObject` and `ReconnectLDAPObject` entries show up fine with current master, but not on the published docs.</s>

Fixes: https://github.com/python-ldap/python-ldap/issues/256

### Clarify and improve documentation of ldap.ldapobject.LDAPObject

Originally, `ldap.ldapobject.LDAPObject` was meant as a read/write alias of the "default" connection class.
However, setting module attributes is bad practice (action at distance): it affects all users of python-ldap, some of which might not be aware of each other.

Clarify that:
* `initialize()` is a thin wrapper around calling `LDAPObject()`,
* it is fine and recommnended to instantiate LDAPObject-like classes directly, and
* it is possible, but not recommended, to set `ldap.ldapobject.LDAPObject` to another class.

Fixes: https://github.com/python-ldap/python-ldap/issues/205

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -32,7 +32,7 @@ This module defines the following functions:
 .. py:function:: initialize(uri [, trace_level=0 [, trace_file=sys.stdout [, trace_stack_limit=None, [bytes_mode=None, [bytes_strictness=None]]]]]) -> LDAPObject object
 
    Initializes a new connection object for accessing the given LDAP server,
-   and return an LDAP object (see :ref:`ldap-objects`) used to perform operations
+   and return an :class:`~ldap.ldapobject.LDAPObject` used to perform operations
    on that server.
 
    The *uri* parameter may be a comma- or whitespace-separated list of URIs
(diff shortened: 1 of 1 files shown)


encukou:
Added another commit to fix https://github.com/python-ldap/python-ldap/issues/205:

> Originally, `ldap.ldapobject.LDAPObject` was meant as a read/write alias of the "default" connection class.
    However, setting module attributes is bad practice (action at distance):    it affects all users of python-ldap, some of which might not be aware    of each other.
>    
> Clarify that:
    * `initialize()` is a thin wrapper around calling `LDAPObject()`,
    * it is fine and recommnended to instantiate LDAPObject-like classes  directly, and
    * it is possible, but not recommended, to set `ldap.ldapobject.LDAPObject` to another class.

encukou:
Merging without formal review, as this is not a code change.

PR #262, opened 2019-01-31
Store message id for exceptions raised from result4()

krisztian-kovacs:
Otherwise calling result4() can return exceptions like NO_SUCH_OBJECT or
COMPARE_TRUE without the caller being able to find the matching asynchronous
operation that failed.

This change adds the message id as an argument to the exception.

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -329,6 +329,14 @@ The module defines the following exceptions:
    is set to a truncated form of the name provided or alias dereferenced
    for the lowest entry (object or alias) that was matched.
 
+   For use in asynchronous operations an optional field :py:const:`msg_id` is
+   also set in the dictionary in cases where the exception can be associated
+   with a request. This can be used in asynchronous code where
+   :py:meth:`result()` returns an exception that is effectively the result of a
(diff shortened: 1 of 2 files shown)


krisztian-kovacs:
Is there any chance of getting this pull request merged? We're using this in production and are forced to use a patched build of python-ldap because of this.

I was sorry to see that a new version has been released without any of the pull requests opened for the project being merged. I think this is really discouraging for 3rd party contributors...

encukou:
Indeed, there's a shortage of contributor time. Sorry for that!
Would you like to help by doing reviews of the outstanding pull requests?

The last release had quite a few pull requests merged. Sorry that yours was one of the ones that didn't make it – when the critical bugfix was reviewed, I took everything that was merged until then.

encukou:
> Indeed, there's a shortage of contributor time.

Still the case.

---

This PR is missing tests and documentation. Would you care to add them?

krisztian-kovacs:
I've updated the patch with docs and tests. Also, I've found out that this would probably fix #177.

krisztian-kovacs:
Thanks for finally merging a solution for this.
I'll update our code to use `msgid` once python-ldap 3.3 is released.

PR #263, opened 2019-02-12
fix "not all arguments converted during string formatting" When Resul…

arpagon:
fix "not all arguments converted during string formatting" When ldap_res Result is a tuple
I tested whit FreeIPA

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -526,7 +526,7 @@ class SimpleLDAPObject:
     except ldap.COMPARE_FALSE:
       return False
     raise ldap.PROTOCOL_ERROR(
-        'Compare operation returned wrong result: %r' % (ldap_res)
+        'Compare operation returned wrong result: %r' % repr(ldap_res)
     )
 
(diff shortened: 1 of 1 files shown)


Lib/ldap/ldapobject.py
```diff
@@ -526,7 +526,7 @@ def compare_ext_s(self,dn,attr,value,serverctrls=None,clientctrls=None):
     except ldap.COMPARE_FALSE:
       return False
     raise ldap.PROTOCOL_ERROR(
-        'Compare operation returned wrong result: %r' % (ldap_res)
+        'Compare operation returned wrong result: %r' % repr(ldap_res)
```
encukou: This will do the `repr` operation twice – once with `repr`, and then for the `%r`.
`print('%r' % repr((1, 2)))` gives `'(1, 2)'` – note the extra quotes.

Could you try this instead?

```suggestion
        'Compare operation returned wrong result: ' + repr(ldap_res)
```

encukou:
Fixed in https://github.com/python-ldap/python-ldap/pull/271

PR #265, opened 2019-03-12
Update the changelog and release 3.2.0

encukou:
With https://github.com/python-ldap/python-ldap/issues/226 fixed, it makes sense to do release.
We have a new feature, X-ORIGIN support, so the release should be 3.2.0.


encukou:
I plan to merge & release tomorrow (even if there's no review – this is not a code change).
If you'd like more time for review, let me know.

PR #267, opened 2019-03-16
Reconnect also on ldap.UNAVAILABLE

spaceone:
LDAP connections may also raise UNAVAILABLE not only SERVER_DOWN. It would be nice if the reconnection class also reconnects in this case.

Below is an example script which can provoke the scenario:

```
import time
import threading
import subprocess

def foo():
    lo = ...  # initialize here!
    lo._retry_max = 10E4
    lo._retry_delay = .001
    x = 0
    lo.search_ext_s("dc=base", ldap.SCOPE_SUBTREE, "uid=Administrator", attr=["uid"])
    s = time.time()
    print("go")
    while (time.time() - s) < run_time:
        x += 1
        lo.search(filter="uid=Administrator", attr=["uid"])
    print("Searches per sec: {}".format(x / run_time))


thread_count = 100
run_time = 10.0
my_thread = [None] * thread_count
for i in range(0, thread_count):
    my_thread[i] = threading.Thread(target=foo)
for t in my_thread:
    t.start()
print("warmup")
time.sleep(3)
print("restart slapd")
subprocess.check_call(["systemctl", "restart", "slapd.service"])
print("restarted")
for t in my_thread:
    t.join()

print("done")
```


spaceone:
The same applies to `ldap.CONNECT_ERROR`. I can adjust the PR if you agree to add also this.

spaceone:
Also `ldap.TIMEOUT` and `ldap.TIMELIMIT_EXCEEDED`.

spaceone:
I adjusted the Pull Request so that one can specify which exceptions should be catched for reconnecting. Therefore it would be extensible and if you want backwards compatible.

If you plan to merge this tell me, I would squash the first and second commit.

spaceone:
@tiran @encukou Any comment on this?

spaceone:
There is another issue, which is triggered, when re-using a connection after a `ldap.SERVER_DOWN` connection. Then the reconnection does not apply as well:

```python
import ldap
import subprocess

lo = ldap.ldapobject.ReconnectLDAPObject(uri)
dn = lo.whoami_s()[3:]

# stop open ldap
subprocess.call(['service', 'slapd', 'stop'])

# do a search, wait for the timeout, handle SERVER_DOWN exception
try:
    print(lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)'))
except ldap.SERVER_DOWN:
    pass  # some handling here

# start open ldap again
subprocess.call(['service', 'slapd', 'start'])

# try to use the connection again
print(lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)'))
# this now raises ldap.INSUFFICIENT_ACCESS, as it is now a unbound connection and does everything anonymously.
```

encukou:
Is it possible to test this?
The existing tests fail, and need to be adjusted.

spaceone:
@tiran Ok, I implemented the 2 new test cases. I would revert making `reconnection_exceptions` configurable, if you agree. I consider it bad style now. We should just add all exceptions which may happen. I wanted to add the parameter so that if we encounter more exceptions in the future, e.g. in different ldap server implementation that it is easier to fix the behavior instead of waiting for a new python-ldap release.

And then I would revert the changes once so that we see that TravisCI actually really fails with the new test cases.

spaceone:
@tiran ping.

mistotebe:
Yes, `CONNECT_ERROR` should definitely be handled, I can see why it could be appropriate for `UNAVAILABLE` as well.

I have some reservations to handling `TIMELIMIT_EXCEEDED` in this way (and to some extent `TIMEOUT` too). The original server is still likely to be contactable, so libldap will use the first URL in the list and you're likely to get the same error back, making no progress?

Not sure what to do when the first server in the list is contactable but broken (TLS, ...), libldap might not take that as a clue to try the next one so we'd need to shuffle the URI list manually?

Regarding the test cases, can't see how they're going to fail, especially the threaded one?

spaceone:
The exceptions come from real-life customer environments where they occurred.
IIRC the test case uses threading to trigger the exception `UNAVAILABLE` with openldap reliable and faster.

mistotebe:
> The exceptions come from real-life customer environments where they occurred.

Please describe a situation where python-ldap has enough information to retry `TIMELIMIT_EXCEEDED` exceptions in a useful way (and why it wouldn't end up reconnecting to the same host for that matter).

> IIRC the test case uses threading to trigger the exception `UNAVAILABLE` with openldap reliable and faster.

The point is that `test_107_reconnect_restore` doesn't seem to check for anything so AFAIK the test will still pass even if the thread fails in any way whatsoever?

spaceone:
> > The exceptions come from real-life customer environments where they occurred.
> 
> Please describe a situation where python-ldap has enough information to retry `TIMELIMIT_EXCEEDED` exceptions in a useful way (and why it wouldn't end up reconnecting to the same host for that matter).
I can't explain `why` this happened on those Systems. I can only provide the information `that` it happened multiple times against OpenLDAP 2.4.

> > IIRC the test case uses threading to trigger the exception `UNAVAILABLE` with openldap reliable and faster.
> 
> The point is that `test_107_reconnect_restore` doesn't seem to check for anything so AFAIK the test will still pass even if the thread fails in any way whatsoever?
But the test will fail if you drop the exception handling.
I am unsure how I can add a test condition there. Maybe I have to mock the `reconnect()` method and make sure inside the exception is raised?

spaceone:
@mistotebe ah OK, now I get you.
So by default we cannot add `TIMELIMIT_EXCEEDED`. I just dropped it for now.

Adding a dependency on https://github.com/jd/tenacity is probably not a good idea now. I like that pyhton-ldap has few dependencies.  

I reworked the tests so that the failing condition is explicit - I thought a failing thread would propagate the error, sorry for that.

mistotebe:
> Adding a dependency on https://github.com/jd/tenacity is probably not a good idea now. I like that python-ldap has few dependencies.

I know, would be interesting if we could provide a way to decide whether (and when) to retry/reconnect in user code, as a hook (which could be tenacity-compatible) or otherwise. That way people could retry whatever they wanted, even for async ops if they really cared. Obviously not now but in a future version/PR.

> I reworked the tests so that the failing condition is explicit - I thought a failing thread would propagate the error, sorry for that.

Looks better, could you simplify the code (removing the "profiling" aspects) and drop the extra print statements too?

Lib/ldap/ldapobject.py
```diff
@@ -833,6 +832,10 @@ class ReconnectLDAPObject(SimpleLDAPObject):
   * retry_delay: specifies the time in seconds between reconnect attempts.
 
   This class also implements the pickle protocol.
+
+  .. versionadded:: 4.0
+    The exceptions :py:exc:`ldap.UNAVAILABLE`, :py:exc:`ldap.CONNECT_ERROR`,
+    :py:exc:`ldap.TIMEOUT` now also trigger a reconnect.
```
mistotebe, line 838: Could you also document this is now configurable?

Lib/ldap/ldapobject.py
```diff
@@ -842,11 +845,12 @@ class ReconnectLDAPObject(SimpleLDAPObject):
     '_reconnect_lock',
     '_last_bind',
   }
+  _reconnect_exceptions = (ldap.SERVER_DOWN, ldap.UNAVAILABLE, ldap.CONNECT_ERROR, ldap.TIMEOUT)
 
   def __init__(
     self,uri,
     trace_level=0,trace_file=None,trace_stack_limit=5,bytes_mode=None,
-    bytes_strictness=None, retry_max=1, retry_delay=60.0, fileno=None
+    bytes_strictness=None, retry_max=1, retry_delay=60.0, fileno=None,
```
mistotebe, line 853: When rebasing, could you also strip the style-only changes from the PR?

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,61 @@ def test105_reconnect_restore(self):
...
+        lo.simple_bind_s(bind_dn, 'user1_pw')
+
+        dn = lo.whoami_s()[3:]
+
+        self.server._proc.terminate()
+        self.server.wait()
+
+        # do a search, wait for the timeout, handle SERVER_DOWN exception
+        with self.assertRaises(ldap.SERVER_DOWN):
+            print(lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)'))
```
mistotebe, line 661: You are still printing here, and below.

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,61 @@ def test105_reconnect_restore(self):
...
+
+        self.server._start_slapd()
+
+        # try to use the connection again
+        print(lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)'))
+        # this used to raise ldap.INSUFFICIENT_ACCESS, as it is became a unbound connection and did everything anonymously.
+
+    def test_107_reconnect_restore(self):
+        excs = []
```
mistotebe, line 670: Could you put a docstring here describing the intent behind this test so that someone can reimplement the busy loop of 100 threads with more intelligent thread coordination/asyncio in the future?

Lib/ldap/ldapobject.py
mistotebe: Hmm, or how about just updating the docstring to reflect reality and a simple outline of the change here?

SvenMW:
I'll try to fix some of the failing tests

Tests/t_ldapobject.py
spaceone: @mistotebe OK, added a docstring. Can you check if suitable?

Lib/ldap/ldapobject.py
mistotebe: Also `SERVER_DOWN` is missing from this list now.

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,66 @@ def test105_reconnect_restore(self):
...
+            lo._retry_delay = .001
+            lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, "cn=user1", attrlist=["cn"])
+            s = time.time()
+            while (time.time() - s) < run_time:
+                lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, filterstr="cn=user1", attrlist=["cn"])
+
+        def reconnect_search_thread():
+            try:
+                _reconnect_search_thread()
+            except Exception as exc:  # ldap.UNAVAILABLE
```
mistotebe, line 692: If `UNAVAILABLE` isn't being raised anymore, this comment is no longer valid.

Lib/ldap/ldapobject.py
```diff
@@ -833,6 +832,11 @@ class ReconnectLDAPObject(SimpleLDAPObject):
   * retry_delay: specifies the time in seconds between reconnect attempts.
 
   This class also implements the pickle protocol.
+
+  .. versionadded:: 4.0
```
mistotebe, line 836: Any reasons on why this couldn't go into 3.5? Doesn't look to be breaking any plausible use cases?

artificial-intelligence:
any update on this? this blocks https://bugs.launchpad.net/keystone/+bug/1953627

Lib/ldap/ldapobject.py
spaceone: don't know - your decidion. for other changes I was asked to put it into 4.0 and didn't thought that 3.5 is still developed.

Tests/t_ldapobject.py
spaceone: it was meant to tell what would be expected if this tests case was broken. i can remove it, if you want

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,66 @@ def test105_reconnect_restore(self):
...
+
+        # do a search, wait for the timeout, handle SERVER_DOWN exception
+        with self.assertRaises(ldap.SERVER_DOWN):
+            lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)')
+
+        self.server._start_slapd()
+
+        # try to use the connection again
+        lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)')
+        # this used to raise ldap.INSUFFICIENT_ACCESS, as it is became a unbound connection and did everything anonymously.
```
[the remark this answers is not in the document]
spaceone: hm but we experienced this behavior. I think this depends on the openldap server configuration for anonymous binds?!

Lib/ldap/ldapobject.py
spaceone: adjusted.

spaceone:
> @spaceone hi! Could you please rebase the commit in this PR? (and possibly check out the above questions), and we'll rerun the tests.

@droideck done. Tests pass.

Tests/t_ldapobject.py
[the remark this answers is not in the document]
mistotebe: In the python-ldap test suite you are in full control of the server config and as such its behaviour for bound/anonymous connections. There should be no "maybe"?

Tests/t_ldapobject.py
mistotebe: If you're not expecting a specific Exception as part of a successful test, then raising one on failure fails the test, correct? No need to catch and retrigger just for the sake of it.

Lib/ldap/ldapobject.py
mistotebe: This is considered a fix, not behaviour change, correct? Then I'd think it qualifies for 3.x, unless @droideck disagrees?

droideck: I'd say it's a minor behaviour change (we stated previously in the docs that we reconnect only on `ldap.SERVER_DOWN`, but now it's `ldap.SERVER_DOWN,ldap.UNAVAILABLE, ldap.CONNECT_ERROR, ldap.TIMEOUT`)

But I think it's really minor and we can release it as a bugfix. So yep, I fully agree with putting it to 3.5!

spaceone: OK, changed to 3.5

Tests/t_ldapobject.py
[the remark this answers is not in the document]
mistotebe: Then you should refer to a ticket? Or better also put that into the docstring like in other tests.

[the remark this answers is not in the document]
spaceone: OK, moved the description of the test to the docstring.

spaceone:
@mistotebe @tiran ping. anything else?

mistotebe:
It is good from my side, @tiran you are marked as "requested changes", could you review again and adjust if appropriate?

mistotebe (approved):
LGTM now

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,72 @@ def test105_reconnect_restore(self):
...
+        After starting the server the LDAP connection can be re-used again as it will reconnect on the next operation.
+        Prior to fixing PR !267 the connection was reestablished but no `bind()` was done resulting in a anonymous search which caused `INSUFFICIENT_ACCESS` when anonymous seach is disallowed.
+        """
+        lo = self.ldap_object_class(self.server.ldap_uri, retry_max=2, retry_delay=1)
+        bind_dn = 'cn=user1,' + self.server.suffix
+        lo.simple_bind_s(bind_dn, 'user1_pw')
+
+        dn = lo.whoami_s()[3:]
+
+        self.server._proc.terminate()
```
tiran, line 662: I don't like the fact that you are using internal APIs of `SlapdObject`. If you need additional features for the test, then you should rather extend the API of `SlapdObject`. How about `SlapdObject.terminate()` and `SlapdObject.resume()`?

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,72 @@ def test105_reconnect_restore(self):
...
+        self.server.wait()
+
+        # do a search, wait for the timeout, ignore SERVER_DOWN
+        with self.assertRaises(ldap.SERVER_DOWN):
+            lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)')
+
+        self.server._start_slapd()
```
tiran, line 669: Private API

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,72 @@ def test105_reconnect_restore(self):
...
+        lo.search_s(dn, ldap.SCOPE_BASE, '(objectClass=*)')
+
+    def test107_reconnect_restore(self):
+        """
+        The idea of this test is to restart the LDAP-Server while there are ongoing searches.
+        This causes :class:`ldap.UNAVAILABLE` to be raised (with |OpenLDAP|) for a short time.
+        To increase the chance of triggering this bug we are starting multiple threads
+        with a large number of retry attempts in a short amount of time.
+        """
+        # TODO: refactor this test
```
tiran, line 681: What's with the TODO? How should the test be refactored?

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,72 @@ def test105_reconnect_restore(self):
...
+            lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, "cn=user1", attrlist=["cn"])
+            s = time.time()
+            while (time.time() - s) < run_time:
+                lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, filterstr="cn=user1", attrlist=["cn"])
+
+        def reconnect_search_thread():
+            try:
+                _reconnect_search_thread()
+            except Exception as exc:
+                excs.append(exc)
```
tiran, line 699: Exceptions have a reference to their traceback, which contains references to their scopes. This creates reference cycles, that mess up reference leak and memory leak checks. You either need to break the reference cycle or not store exception objects.

Tests/t_ldapobject.py
```diff
@@ -643,6 +646,72 @@ def test105_reconnect_restore(self):
...
+
+        thread_count = 100
+        run_time = 10.0
+        my_thread = [None] * thread_count
+        for i in range(0, thread_count):
+            my_thread[i] = threading.Thread(target=reconnect_search_thread)
+        for t in my_thread:
+            t.start()
+        time.sleep(3)
```
tiran, line 708: Magic sleeps are problematic. On one hand they tend to break on slow or busy machines like CI and build systems. On the other hand they slow down fast machines. My dev box can run the tests in 10 seconds. This magic sleep would slow down tests by 30%.

I suggest that you use a proper synchronization primitive like a barrier. Also 100 threads seems a bit excessive. 10 threads are probably enough.

tiran (changes requested):

spaceone:
Sorry, I don't have the time to change the test case according to those requirements.

spaceone:
OK, the tests have been refactored and run:

1. Fixed Code:
```
$ PYTHONPATH=Lib/ python3 -m unittest -v Tests.t_ldapobject.Test01_ReconnectLDAPObject.test107_reconnect_restore
test107_reconnect_restore (Tests.t_ldapobject.Test01_ReconnectLDAPObject.test107_reconnect_restore)
The idea of this test is to restart the LDAP-Server while there are ongoing searches. ... ok

----------------------------------------------------------------------
Ran 1 test in 10.312s

OK

```

2. Failing test shows that it's currently broken:

```
$ python3 -m unittest -v Tests.t_ldapobject.Test01_ReconnectLDAPObject.test107_reconnect_restore                                                                                        
test107_reconnect_restore (Tests.t_ldapobject.Test01_ReconnectLDAPObject.test107_reconnect_restore)                    
The idea of this test is to restart the LDAP-Server while there are ongoing searches. ... Exception occurred {'result': 52, 'desc': 'Server is unavailable', 'ctrls': []} Traceback (most recent call last):                                   
  File "/home/fbest/git/python-ldap/Tests/t_ldapobject.py", line 701, in reconnect_search_thread                       
    _reconnect_search_thread()                                                                                         
  File "/home/fbest/git/python-ldap/Tests/t_ldapobject.py", line 697, in _reconnect_search_thread
    lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, filterstr="cn=user1", attrlist=["cn"])
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 1030, in search_ext_s
    return self._apply_method_s(SimpleLDAPObject.search_ext_s,*args,**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 968, in _apply_method_s
    return func(self,*args,**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 625, in search_ext_s
    return self.result(msgid,all=1,timeout=timeout)[1]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 535, in result
    resp_type, resp_data, resp_msgid = self.result2(msgid,all,timeout)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 539, in result2
    resp_type, resp_data, resp_msgid, resp_ctrls = self.result3(msgid,all,timeout)
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 543, in result3
    resp_type, resp_data, resp_msgid, decoded_resp_ctrls, retoid, retval = self.result4(
                                                                           ^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 553, in result4
    ldap_result = self._ldap_call(self._l.result4,msgid,all,timeout,add_ctrls,add_intermediates,add_extop)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 128, in _ldap_call
    result = func(*args,**kwargs)
             ^^^^^^^^^^^^^^^^^^^^
ldap.UNAVAILABLE: {'result': 52, 'desc': 'Server is unavailable', 'ctrls': []}
…
                                                                                                                       
FAIL                                                                                                                   
                                                           
======================================================================                           
FAIL: test107_reconnect_restore (Tests.t_ldapobject.Test01_ReconnectLDAPObject.test107_reconnect_restore)
The idea of this test is to restart the LDAP-Server while there are ongoing searches. 
----------------------------------------------------------------------       
  File "/home/fbest/git/python-ldap/Tests/t_ldapobject.py", line 717, in test107_reconnect_restore
    self.assertEqual(excs, [])      
AssertionError: Lists differ: [("{'result': 52, 'desc': 'Server is unava[6916 chars]\n')] != []
                                                                                                                       
First list contains 3 additional elements.            
First extra element 0:                             
("{'result': 52, 'desc': 'Server is unavailable', 'ctrls': []}", 'Traceback (most recent call last):\n  File "/home/fbest/git/python-ldap/Tests/t_ldapobject.py", line 701, in reconnect_search_thread\n    _reconnect_search_thread()\n  File 
"/home/fbest/git/python-ldap/Tests/t_ldapobject.py", line 697, in _reconnect_search_thread\n    lo.search_ext_s(self.server.suffix, ldap.SCOPE_SUBTREE, filterstr="cn=user1", attrlist=["cn"])\n  File "/usr/lib/python3/dist-packages/ldap/lda
pobject.py", line 1030, in search_ext_s\n    return self._apply_method_s(SimpleLDAPObject.search_ext_s,*args,**kwargs)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/l
dap/ldapobject.py", line 968, in _apply_method_s\n    return func(self,*args,**kwargs)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 625, in search_ext_s\n    return self.result(msg
id,all=1,timeout=timeout)[1]\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 535, in result\n    resp_type, resp_data, resp_msgid = self.result2(msgid,all,timeout)\n   
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 539, in result2\n    resp_type, resp_data, resp_msgid, resp_ctrls = self.result3(msgid,all,timeout)\n    
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 543, in result3\n    resp_type, resp_data, resp_msgid, decoded_resp_ctrls, retoid, retval = se
lf.result4(\n                                                                           ^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 553, in result4\n    ldap_result = self._ldap_call(self._l.result4,msgi
d,all,timeout,add_ctrls,add_intermediates,add_extop)\n                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/usr/lib/python3/dist-packages/ldap/ldapobject.py", line 128, in _lda
p_call\n    result = func(*args,**kwargs)\n             ^^^^^^^^^^^^^^^^^^^^\nldap.UNAVAILABLE: {\'result\': 52, \'desc\': \'Server is unavailable\', \'ctrls\': []}\n')
                                                                                                                       
Diff is 7994 characters long. Set self.maxDiff to None to see it.                                         

----------------------------------------------------------------------
Ran 1 test in 10.318s

FAILED (failures=1)
```

Tests/t_ldapobject.py
spaceone: Thanks, that's very much what I searched for to make the test clean.

Tests/t_ldapobject.py
spaceone: Test is refactored according to your threading.Barrier suggestion.

droideck:
@spaceone I think we are close to merging.
Could you please rebase the code once again, so we can check the tests?

PR #274, opened 2019-03-31
 Add ldap.dn.compare_dn() to compare equality of distinguished names 

spaceone:
Added a function which allows to compare two DN's for equality.

To be discussed: the third commit makes it possible to also compare the attribute values of a DN (e.g. "uid=Foo" == "uid=foo"). But it requires some initialization, as it depends on the actual used schema on the LDAP server.

### DIFF
--- a/Lib/ldap/dn.py
+++ b/Lib/ldap/dn.py
@@ -126,4 +126,27 @@ def compare_dn(a, b):
     """
     Compares two distinguished names and return True if they are equal.
     """
-    return [sorted((x.lower(), y, z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(a)] == [sorted((x.lower(), y, z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(b)]
+    case_insensitive_attributes = [attr.lower() for attr in compare_dn.case_insensitive_attributes]
+
+    def normalize(attr, value):
+        if attr.lower() in case_insensitive_attributes:
+            value = value.lower()
+        return value
+    return [sorted((x.lower(), normalize(x, y), z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(a)] == [sorted((x.lower(), normalize(x, y), z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(b)]
+
+
+compare_dn.case_insensitive_attributes = []
+
+
+def initialize_case_insensitive_attributes(lo):
+    """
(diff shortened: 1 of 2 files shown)


encukou:
To be correct, DN comparison would have to take into account not only schema, but also *any* equality matching rules – not just `caseIgnore`. We should not attempt to guess unless we have the correct answer (according to both standards and actual usage).
Also, don't initialize and depend on global state – python-ldap can manage many connections with different schemas. The `compare_dn` function could, for example, take a schema as argument.

An "approximate" implementation can still be valuable, but it's out of python-ldap's scope. It could be added in a more specialized library built on top of python-ldap.

Lib/ldap/dn.py
```diff
@@ -120,3 +120,33 @@ def is_dn(s,flags=0):
...
+    case_insensitive_attributes = [attr.lower() for attr in compare_dn.case_insensitive_attributes]
+
+    def normalize(attr, value):
+        if attr.lower() in case_insensitive_attributes:
+            value = value.lower()
+        return value
+    return [sorted((x.lower(), normalize(x, y), z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(a)] == [sorted((x.lower(), normalize(x, y), z) for x, y, z in rdn) for rdn in ldap.dn.str2dn(b)]
+
+
+compare_dn.case_insensitive_attributes = []
```
tiran: This global cache assumes that an application only contacts one LDAP server or only LDAP servers with compatible schema. The cache is invalid in case of incompatible schema.

Lib/ldap/dn.py
```diff
@@ -120,3 +120,33 @@ def is_dn(s,flags=0):
...
+
+def compare_dn(a, b):
+    """
+    Compares two distinguished names and return True if they are equal.
+    """
+    case_insensitive_attributes = [attr.lower() for attr in compare_dn.case_insensitive_attributes]
+
+    def normalize(attr, value):
+        if attr.lower() in case_insensitive_attributes:
+            value = value.lower()
```
tiran: This assumes trivial lower case conversation. LDAP servers can have more complex and language specific conversation and comparsion.

spaceone:
Are you aware of any C function which could be exposed for this purpose?

encukou:
> Are you aware of any C function which could be exposed for this purpose?

No, unfortunately. This needs to be handled on the server.

---

This would be suitable for a library that builds *on top* of python-ldap, and which has more information about the schema or data. Python-ldap itself is too low-level.
Please re-open if you want to discuss further.

PR #277, opened 2019-04-12
Make tests pass after 2028

bmwiedemann:
Extend the lifetime of test certs to not expire so soon.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.

See https://reproducible-builds.org/ for why this matters.

I tested, that with the patch, everything passed in 2034.


tiran:
Thanks for the pull request. I like the idea to extend the certs. However 2994 seems a bit excessive. For now I'd prefer to keep dates before 2038 to avoid Epochalypse bugs. How about a validity of 15 years, then?

bmwiedemann:
If there were y2038 bugs, it would be better to detect and fix them now rather than in 15y.
At least these certs do not trigger any test failures on i586 or x86_64 neither today nor in 2037.
So should be fine.

bmwiedemann:
any concerns that prevent merging?

bmwiedemann:
ping. I'd still like to see this merged

Issue #280, opened 2019-04-26
ldapurl parsing fails when scheme is uppercase


dmulder:
https://github.com/python-ldap/python-ldap/blob/adf9dc402380572b1d3b5bc6fa1d353bbed35074/Lib/ldapurl.py#L260

This only checks lower case instances of ldap, ldaps, and ldapi. What if they are upper case?

tiran:
RFC 1738 defines the schema of an uniformed resource locator as a sequence of characters from the set ``a-z0-9+.-``. ``LDAP://host`` is not a valid URL schema identifier. RFC 1738 also suggests that applications should treat upper case schema like lower case schema. The RFC uses the word *should*, not *must*, so it's not a requirement but an optional feature.

I'm -0 on the change and let Petr decide. I have been working with LDAP for almost a decade and never ran into a situation with an upper case scheme. The change also needs tests and a thorough review of the code base.

encukou:
I'm +0 – this is very unlikely to break existing code, and *should* features are nice to have.

quanah:
Reading [RFC2119](https://www.ietf.org/rfc/rfc2119.txt), I'm not entirely sure I agree with your definition of should.

quanah:
I would also note that OpenLDAP (which strives for being as RFC compliant as possible) accepts uppercase URI schemes:

```
ldapwhoami -x -H LDAP:// -D "cn=Marice McCaugherty,ou=Product Testing,dc=example,dc=com" -w secret
dn:cn=Marice McCaugherty,ou=Product Testing,dc=example,dc=com
```

This would generally indicate to me that it is more correct to accept both upper and lower case format for the URI scheme.

PR #281, opened 2019-04-26
Fix ldapurl parsing fails when scheme is uppercase

dmulder:
Fixes issue mentioned in #280.


encukou:
I'm not opposed to moving forward. But yes, this does need tests.

dmulder:
@tiran do you have docs for the python-ldap test suite? I've never written a test for this project before.

tiran:
> @tiran do you have docs for the python-ldap test suite? I've never written a test for this project before.

The tests are in ``Tests/`` directory. You can add some upper-case examples to this test class:

https://github.com/python-ldap/python-ldap/blob/7e084aec1ba9ced25b44fd3db77e65242a827806/Tests/t_ldapurl.py#L31-L53

tiran:
I have opened PR https://github.com/python-ldap/python-ldap/pull/347 for you to get the fix into 3.3.0.

PR #283, opened 2019-05-02
Get rid of expected failures in tokenizer tests

mistotebe:
One of the failures actually stems from using the wrong method. Extending the tokenizer to deal with escapes gets rid of the other.


encukou:
I am out of my depth here: the details of LDAP tokenization and Oracle Internet Directory escape me :(
@tiran, do you want to review this?

Issue #61, opened 2017-11-30
Use PyStructSequence for result() return value


encukou:
The [result](http://python-ldap.readthedocs.io/en/latest/reference/ldap.html#ldap.LDAPObject.result)/[result2](http://python-ldap.readthedocs.io/en/latest/reference/ldap.html#ldap.LDAPObject.result2)/[result3](http://python-ldap.readthedocs.io/en/latest/reference/ldap.html#ldap.LDAPObject.result3)/[result4](http://python-ldap.readthedocs.io/en/latest/reference/ldap.html#ldap.LDAPObject.result4) API looks quite silly by today's standards, if you don't mind me saying.

Do what [os.stat does](https://docs.python.org/3/library/os.html#os.stat_result). `result()` should return an object that:
- for backwards compatibility, acts like a tuple of the first two elements
- has everything accessible as attributes.

Such a type can be defined at the C level using [Struct Sequence](https://docs.python.org/3/c-api/tuple.html#struct-sequence-objects).
In Python 2 this API is private; use a regular namedtuple subclass?

Deprecate result2/result3/result4 (but keep them around indefinitely).

spaceone:
result4 is definetly used in the real world because it's the only way to get the server control response.

encukou:
Indeed, and `result4()` should stay as it is. The proposed usage for `result()` is something like:

```
result = connection.result()
use(result.type)
use(result.data)
use(result.msgid)
use(result.controls)
use(result.name)
use(result.value)
```

with `type, data = connection.result()` still working, for backwards compatibility

spaceone:
This would be very nice.

PR #288, opened 2019-05-17
Additional exception fields

mistotebe:
Work to improve how we raise exceptions on result, deals with controls and msgid being missing (#177, #208).

Ready for merging (pending #282, #283, #284 have been accepted) apart from last commit which is mainly here for initial review and can be split into its own PR if you want.

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -1086,7 +1086,6 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
     char *retoid = 0;
     PyObject *valuestr = NULL;
     int result = LDAP_SUCCESS;
-    char **refs = NULL;
     LDAPControl **serverctrls = 0;
 
     if (!PyArg_ParseTuple
@@ -1157,23 +1156,15 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
         }
 
         LDAP_BEGIN_ALLOW_THREADS(self);
-        rc = ldap_parse_result(self->ldap, msg, &result, NULL, NULL, &refs,
+        rc = ldap_parse_result(self->ldap, msg, &result, NULL, NULL, NULL,
                                &serverctrls, 0);
         LDAP_END_ALLOW_THREADS(self);
     }
 
     if (result != LDAP_SUCCESS) {       /* result error */
-        char *e, err[1024];
-
-        if (result == LDAP_REFERRAL && refs && refs[0]) {
-            snprintf(err, sizeof(err), "Referral:\n%s", refs[0]);
-            e = err;
-        }
-        else
-            e = "ldap_parse_result";
         ldap_controls_free(serverctrls);
         Py_XDECREF(valuestr);
-        return LDAPraise_for_message(self->ldap, e, msg);
+        return LDAPraise_for_message(self->ldap, "ldap_parse_result", msg);
     }
 
     if (!(pyctrls = LDAPControls_to_List(serverctrls))) {
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -415,7 +415,7 @@ l_ldap_unbind_ext(LDAPObject *self, PyObject *args)
     LDAPControl_List_DEL(client_ldcs);
 
     if (ldaperror != LDAP_SUCCESS)
-        return LDAPerror(self->ldap, "ldap_unbind_ext");
+        return LDAPerror(self->ldap);
 
     self->valid = 0;
     Py_INCREF(Py_None);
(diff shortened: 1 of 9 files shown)


somay:
LGTM regarding #208 .
My patch is already merged well.

mistotebe:
@tiran, @encukou any thoughts on this and the dependent PRs?

encukou:
This is out of my depth.
@tiran, do you want to review this?

mistotebe:
Any ideas on a review? Without this it's impossible to have more than one operation pending on a connection.

encukou:
My review will be sub-par, but it looks like I'll need to do it to move this forward. I'll dedicate a day next week to python-ldap.

mistotebe:
Thanks, I'll rebase it in the meantime.

mistotebe:
Only fails on the same doctest as master now.

encukou:
The last commit should go into a separate PR:
* Adding a new item to the `result4`'s return value tuple will break code that unpacks the tuple. This needs either adding a `result5()`, or fixing #61.
* The `raise_for_result` and `RAISE_*` and need documentation.

Modules/constants.c
```diff
@@ -48,38 +49,81 @@ LDAPerr(int errnum)
...
+            pyresult = PyInt_FromLong(msgid);
+            if (pyresult)
+                PyDict_SetItemString(info, "msgid", pyresult);
+            Py_XDECREF(pyresult);
+        }
+
+        pyresult = PyInt_FromLong(errnum);
+        if (pyresult)
+            PyDict_SetItemString(info, "result", pyresult);
+        Py_XDECREF(pyresult);
```
encukou, line 126: This is missing error handling; `PyInt_FromLong` and `PyDict_SetItemString` can fail.
Same for other Python C-API below.

Modules/LDAPObject.c
```diff
@@ -1172,9 +1172,8 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
         else
             e = "ldap_parse_result";
         ldap_controls_free(serverctrls);
-        ldap_msgfree(msg);
```
encukou, line 1175: Since this `ldap_msgfree` is removed, it looks like `msg` will leak.

Modules/constants.c
```diff
@@ -48,38 +49,81 @@ LDAPerr(int errnum)
 
 /* Convert an LDAP error into an informative python exception */
 PyObject *
-LDAPerror(LDAP *l, char *msg)
+LDAPraise_for_message(LDAP *l, char *msg, LDAPMessage *m)
```
encukou, line 52: The `msg` argument is unused, can it be removed?

Modules/LDAPObject.c
```diff
@@ -1193,27 +1193,19 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
...
-    }
-
     if (pmsg == NULL) {
         retval = NULL;
     }
     else {
         /* s handles NULL, but O does not */
         if (add_extop) {
-            retval = Py_BuildValue("(OOiOsO)", result_str, pmsg, res_msgid,
+            retval = Py_BuildValue("(iOiOsO)", res_type, pmsg, res_msgid,
```
encukou: This changes the return type for all the `result` functions, right?

Modules/LDAPObject.c
mistotebe: It is now passed over to `LDAPraise_for_message` which calls `ldap_parse_result(..., 1)` in turn.

Modules/constants.c
mistotebe: Is there an example how this should be done? All I can find is both of them treated the this way across the codebase.

Modules/LDAPObject.c
mistotebe: No change in type here, at this point `result_str` is always positive so we never even returned `None`. Then it might as well be converted to `int()` in `Py_BuildValue` rather than manually...

Modules/constants.c
mistotebe: Thanks for spotting it. That wasn't intentional and should be fixed now.

mistotebe:
I've moved the WIP commit into its own branch, but kept the `.errnum` one in for now. Let me know if it should be moved too.

Modules/constants.c
mistotebe: Ah, it was never used and https://github.com/python-ldap/python-ldap/blob/master/Lib/ldap/ldapobject.py#L325 makes it impossible to actually propagate it without breaking/changing existing tests.

It feels quite wrong to it that way, but I guess it's unavoidable for now.

Modules/constants.c
encukou: You are right. I'll send a separate PR to add error handling, so this isn't blocked.

encukou: For reference, it's here: https://github.com/python-ldap/python-ldap/pull/341

Issue #242, opened 2018-09-25
Retry on EINTR


rgbyrnes:
python-ldap doesn't appear to retry ldap calls that fail with EINTR.

This is fairly easy to reproduce.  If we run a loop doing repeated ldap calls, install a signal handler (that does nothing), and then send the handled signal a few times to the process, it fails almost immediately, raising an exception like ...

```
SERVER_DOWN: {u'info': 'Interrupted system call', 'errno': 4, 'desc': u"Can't contact LDAP server"}
```

It's also fairly easy to fix, since all of the ldap calls seem to happen in `ldap.ldapobject.SimpleLDAPObject._ldap_call`.  If I add a little wrapper like ...

```
from errno import EINTR

def _retry_if_interrupted(func, *args, **kwargs):
  """
  Call func, retrying if it raises an LDAPError with errno == EINTR.
  """
  while True:
    try:
      return func(*args, **kwargs)
    except LDAPError as e:
      try:
        if e.args[0]['errno'] != EINTR:
          raise
      except (AttributeError, IndexError, KeyError):
        raise e
```

... and change the ldap call to ...

```
        result = _retry_if_interrupted(func, *args, **kwargs)
```

... then I can no longer reproduce the issue.

We have observed this as the source of sporadic ldap failures in real apps.  Requiring callers to do their own EINTR retries seems onerous, and raising exceptions for EINTR "failures" doesn't seem useful, hence the suggestion to retry.

I didn't track down where EINTR originates within OpenLDAP.  The OpenLDAP code has many references to EINTR that suggest an intention to retry ... maybe some cases were missed?  Retrying by python-ldap seems consistent with that, at least.

Operating system: Linux, RHEL 7.5 with kernel-3.10.0-862.11.6.el7.x86_64
Python version: 2.7.15
python-ldap version: 3.1.0
OpenLDAP version: 2.4.39

spaceone:
doesn't it work when you use `ldap.ldapobject.ReconnectLDAPObject` instead of `ldap.ldapobject.SimpleLDAPObject`?

Breloc:
I met the same problem, but I've managed to solve it by enabling `OPT_RESTART` with  :
```python
ldap.set_option(ldap.OPT_RESTART, ldap.OPT_ON)
```
[...]

PR #291, opened 2019-05-24
Add retry on EINTR.

rgbyrnes:
Fixes https://github.com/python-ldap/python-ldap/issues/242.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -32,6 +32,22 @@ from ldap.compat import reraise
 
 from ldap import LDAPError
 
+from errno import EINTR
+
+def _retry_if_interrupted(func, *args, **kwargs):
+  """Call func, retrying if it raises an LDAPError with errno == EINTR.
+  """
+
+  while True:
+    try:
+      return func(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


dmulder:
FYI, I was getting this error:
ldap.SERVER_DOWN: {'desc': "Can't contact LDAP server", 'errno': 4, 'info': 'Interrupted system call'}
I think my server really is unreachable, and applying this patch now causes an infinite loop in my code.

rgbyrnes:
I'm surprised that you're seeing consistent, repeated errno == EINTR with ldap.SERVER_DOWN ... why would that happen?  We've been running with this patch for > 9 months and have never seen that.

I suppose it couldn't hurt to terminate the loop after some finite number of retries, but that sorta feels like it's covering up some other issue.  If you can reproduce the behavior, it would be interesting to strace.  Are you using OpenLDAP?

encukou:
@dmulder Which server are you using?

encukou:
No response :(
@rgbyrnes, would you mind terminating the loop? An exception is better than infinite loop – especially since this code has no tests.

encukou:
Still no response :(
Please re-open this if you have any additional information.

Issue #246, opened 2018-10-03
missing generated password from the server on the extended password modify operation


akshayadhikari1:
Issue description: 

I was trying to use passwd_s() for various password change operation, what I observed is there is no way to extract the system generated password when we do not provide new password. I checked the logs it does generate the random system password.

Steps to reproduce: 

passwd_s(USER_DN, OLD_PASSWORD, None)

Operating system:
 
Fedora-28

Python version: 

3.6.6

python-ldap version:

2.4.46-3.fc28

quanah:
This works for me:
```
import ldap
l = ldap.initialize('ldap://localhost')
l.simple_bind_s("dc=example,dc=com", "secret")
msgid = l.passwd_s("cn=Marice McCaugherty,ou=Product Testing,dc=example,dc=com", None, None)
password=msgid[1].decode('latin1').split()[1]
print("New password:%s" % password)
```

I get:

```
# ./password.py
New password:Xjt9leMy
```

And I can confirm this password works:

```
# ldapwhoami -x -H LDAP:// -D "cn=Marice McCaugherty,ou=Product Testing,dc=example,dc=com" -w Xjt9leMy
dn:cn=Marice McCaugherty,ou=Product Testing,dc=example,dc=com
```

PR #299, opened 2019-08-20
Extract new password in passwd_s (#246)

mistotebe:
Care has been taken to preserve backwards compatibility for the regular case of no password being requested too.

### DIFF
--- a/Lib/ldap/extop/__init__.py
+++ b/Lib/ldap/extop/__init__.py
@@ -65,3 +65,4 @@ class ExtendedResponse:
 
 # Import sub-modules
 from ldap.extop.dds import *
+from ldap.extop.passwd import PasswordModifyResponse
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -711,7 +711,7 @@ class Test00_SimpleLDAPObject(SlapdTestCase):
             l.passwd_s(dn, "bogus", "ignored")
 
         # have the server generate a new random pw
-        respoid, respvalue = l.passwd_s(dn, "initial", None)
+        respoid, respvalue = l.passwd_s(dn, "initial", None, extract_newpw=True)
         self.assertEqual(respoid, None)
 
         password = respvalue.genPasswd
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -687,6 +687,45 @@ class Test00_SimpleLDAPObject(SlapdTestCase):
             self._ldap_conn.result()
         self.assertEqual(cm.exception.args[0]["msgid"], msgid)
 
+    def test_passwd_s(self):
+        l = self._ldap_conn
+
+        # first, create a user to change password on
+        dn = "cn=PasswordTest," + self.server.suffix
+        result, pmsg, msgid, ctrls = l.add_ext_s(
+            dn,
+            [
+                ('objectClass', b'person'),
+                ('sn', b'PasswordTest'),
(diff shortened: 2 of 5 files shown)


mistotebe:
So you want me to add a `PASSWD_NEW = object()` into ldap.py?

Yes, [RFC 3062](https://tools.ietf.org/html/rfc3062#section-2.2) says so. Also, there is some confusion on the OPTIONAL status of genPasswd in the ASN.1 spec vs. text spec but I replicated the more permissive version for better interoperability.

mistotebe:
@tiran Updated merge request, just the new_pw handling to resolve then?

mistotebe:
Wouldn't your suggestion make it impossible to ever deprecate the old behaviour?

mistotebe:
@tiran Ok, this might be better? If it is, feel free to squash the relevant commits before merging.

encukou:
@tiran, do you want to finish the review?

encukou:
My review will be sub-par, but it looks like I'll need to do it to move this forward. I'll dedicate a day next week to python-ldap.

encukou:
@mistotebe Overall, this looks good to me!
I've rebased so tests can pass, and made a few nitpicks directly in your branch. Does that look OK?

The `extract_newpw` flag is not ideal API, but it should work for now. When https://github.com/python-ldap/python-ldap/issues/61 is implemented, it can be deprecated.

mistotebe:
Thanks for the changes. I agree the result API needs a complete overhaul, couldn't do that here and frankly, the exception work is a prerequisite to that anyway.

Tests/t_ldapobject.py
```diff
@@ -673,6 +673,45 @@ def test_compare_s_invalidattr(self):
...
+            dn,
+            [
+                ('objectClass', b'person'),
+                ('sn', b'PasswordTest'),
+                ('cn', b'PasswordTest'),
+                ('userPassword', b'initial'),
+            ]
+        )
+        self.assertEqual(result, ldap.RES_ADD)
+        self.assertEqual(type(msgid), type(0))
```
tiran, line 691:
```suggestion
        self.assertIsInstance(msgid, int)
```

Tests/t_ldapobject.py
```diff
@@ -673,6 +673,45 @@ def test_compare_s_invalidattr(self):
...
+        # try changing password with a wrong old-pw
+        with self.assertRaises(ldap.UNWILLING_TO_PERFORM):
+            l.passwd_s(dn, "bogus", "ignored")
+
+        # have the server generate a new random pw
+        respoid, respvalue = l.passwd_s(dn, "initial", None, extract_newpw=True)
+        self.assertEqual(respoid, None)
+
+        password = respvalue.genPasswd
+        self.assertEqual(type(password), bytes)
```
tiran, line 704:
```suggestion
        self.assertIsInstance(password, bytes)
```

Lib/ldap/extop/__init__.py
```diff
@@ -65,3 +65,4 @@ def decodeResponseValue(self,value):
 
 # Import sub-modules
 from ldap.extop.dds import *
+from ldap.extop.passwd import *
```
tiran, line 68:
```suggestion
from ldap.extop.passwd import PasswordModifyResponse
```

quanah:
> For some reason, Travis CI is not reporting results.

You may wish to see if the discussion here is of help: https://mta.openssl.org/pipermail/openssl-project/2020-June/002033.html

PR #303, opened 2019-09-20
ldap.dn.is_dn: Fix typo in docstring

encukou:
Cherry-picked change from https://github.com/python-ldap/python-ldap/pull/272


encukou:
Merging without review, since this is not a code change.

PR #307, opened 2019-10-24
docs: python-tox no longer exists in Debian Buster

eyalch:


encukou:
I assume it'll be the same in Debian Bullseye and newer ones. Right?
I think it would make sense to put `tox` in the instructions and have a note for *older* Debian versions.

Doc/installing.rst
```diff
@@ -148,6 +148,8 @@ Packages for building and testing::
    # apt-get install build-essential python3-dev python2.7-dev \
        libldap2-dev libsasl2-dev slapd ldap-utils python-tox \
```
tiran, line 149:
```suggestion
       libldap2-dev libsasl2-dev slapd ldap-utils tox \
```

Doc/installing.rst
```diff
@@ -148,6 +148,8 @@ Packages for building and testing::
...
        lcov valgrind
+       
+*Note:* Replace ``python-tox`` with ``tox`` in Debian Buster.
```
tiran, line 152:
```suggestion
.. note::
   
   On older releases ``tox`` was called ``python-tox``.
```

tiran:
Travis CI is blocked

PR #310, opened 2019-12-04
fixed bug in ldif module so it works with python 3.x

atterdag:
Fixed bug with python 3.x by ensuring to encode, and decode attr_value before being evaluated.

Tested with Python 3.6.8 and Python 2.7.15+ on Ubuntu 18.04.3 x86_64.

### DIFF
--- a/Lib/ldif.py
+++ b/Lib/ldif.py
@@ -125,7 +125,7 @@ class LDIFWriter:
     of special chars or because attr_type is in self._base64_attrs
     """
     if isinstance(attr_value, str):
-      attr_value = str.encode(attr_value,'ascii')
+      attr_value = str.encode(attr_value,'utf-8')
     return attr_type.lower() in self._base64_attrs or \
            not safe_string_re.search(attr_value) is None
 
--- a/Lib/ldif.py
+++ b/Lib/ldif.py
@@ -124,6 +124,8 @@ class LDIFWriter:
     returns 1 if attr_value has to be base-64 encoded because
     of special chars or because attr_type is in self._base64_attrs
     """
+    if isinstance(attr_value, str):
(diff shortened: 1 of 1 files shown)


tiran:
The entry from the example is wrong. python-ldap uses bytes on Python 3 except when the LDAP protocol mandates UTF-8. Attribute names and DNs are text, attribute values are lists of bytes. This is documented at https://www.python-ldap.org/en/python-ldap-3.2.0/bytes_mode.html

```
>>> import ldif, sys
>>> entry = {'objectClass': [b'top', b'person'], 'cn': [b'Michael Stroeder'],'sn': [b'Stroeder']}
>>> dn = 'cn=Michael Stroeder,ou=Test'
>>> ldif_writer = ldif.LDIFWriter(sys.stdout)
>>> ldif_writer.unparse(dn, entry)
dn: cn=Michael Stroeder,ou=Test
cn: Michael Stroeder
objectClass: top
objectClass: person
sn: Stroeder
```

atterdag:
ok, so before I withdraw this PR, can I suggest that you change the example to something like this:

```
import sys,ldif

def encode_values(entries):
    encoded_dict = {}
    for key,list in entries.items():
        encoded_list = []
        for entry in list:
            encoded_list.append(bytes(entry,'utf-8'))
        encoded_dict[key] = encoded_list
    return encoded_dict

DN = 'cn=Michael Stroeder,ou=Test'
ENTRY = {'objectClass':['top','person'],'cn':['Michael Stroeder'],'sn':['Stroeder']}

ldif_writer=ldif.LDIFWriter(sys.stdout)
ldif_writer.unparse(DN,encode_values(ENTRY))
```

Or should I make a PR to correct the documentation? ;)

Issue #273, opened 2019-03-30
python setup.py build fails on OpenBSD


kwatsen:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

Issue description:

Steps to reproduce: 

Operating system: 

Python version: 

python-ldap version:

kwatsen:
Note: a fix is to add these two lines to th `[_ldap]` section in `setup.py`:
```
+library_dirs = /usr/local/lib
+include_dirs = /usr/local/include /usr/local/include/sasl
```
After adding the above two lines, then it compiles cleanly.

I don't know if there is a way to add this to `python-ldap/Build/build-openbsd`, as I think that file is more for compiling in the OpenBSD `ports` tree (not for, e.g., `pip install`).

Happy to submit a PR if the devs want to suggest a good multi-platform fix.

tiran:
I'm not that familiar with OpenBSD. Isn't there a better way to configure the compiler, linker, and dynamic loader to pick up these locations by default?

Python just calls the compiler and linker. You can also set LDFLAGS and CFLAGS env vars.

kwatsen:
Wait, no, that can't be right.  `pip install` should just work.  Yes, I've debugged it some by cloning the git repo and running `python setup.py ...`, but most wouldn't...

I think the needed fix is in the `setup.py` file, around where it says: `#-- Read the [_ldap] section of setup.cfg`.  My guess is that you want something like `if platform.system() == 'OpenBSD': ...`

Ideally, the details could stay in the setup.cfg file:
```
[_ldap_openbsd]
library_dirs = /usr/local/lib
include_dirs = /usr/local/include /usr/local/include/sasl
```

and then the setup.py just brings this section in when platform is "OpenBSD"?

tiran:
I agree that it should just work but I don't agree with your proposed solution. python-ldap is not the correct place to solve a compiler and linker issue. It's either a problem with OpenBSD, compiler settings, OpenBSD's package manager, or Python on OpenBSD. The build system should pick up default locations for headers and libraries.

The misconfiguration affects all Python packages with a C extension, not just python-ldap.

gegeweb:
Same issue with 3.2.0 on FreeBSD (11.2-RELEASE).
No problem to compile the 3.1.0 in a virtualenv on the same system.

quanah:
[...]

PR #313, opened 2019-12-23
setup.py: Add include_dirs for FreeBSD

iblis17:


encukou:
This will add `/usr/local/include` to the include paths even if a different path is explicitly set in `[_ldap]` section of `setup.cfg`. 

See also https://github.com/python-ldap/python-ldap/issues/273

tiran:
The include path should not be hard-coded. Also see https://github.com/python-ldap/python-ldap/issues/273#issuecomment-634526181

Issue #306, opened 2019-10-23
Example in documentation does not work


robert-sherwood:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

Issue description:
Example at the following link:
https://github.com/python-ldap/python-ldap/blob/master/Doc/reference/ldif.rst#example

>>> import sys,ldif
>>> entry={'objectClass':['top','person'],'cn':['Michael Stroeder'],'sn':['Stroeder']}
>>> dn='cn=Michael Stroeder,ou=Test'
>>> ldif_writer=ldif.LDIFWriter(sys.stdout)
>>> ldif_writer.unparse(dn,entry)

When entering this code in the python console, the following error is received:
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/plainid/PycharmProjects/LDIF_Generator/venv/lib/python3.7/site-packages/ldif.py", line 198, in unparse
    self._unparseEntryRecord(record)
  File "/Users/plainid/PycharmProjects/LDIF_Generator/venv/lib/python3.7/site-packages/ldif.py", line 155, in _unparseEntryRecord
    self._unparseAttrTypeandValue(attr_type,attr_value)
  File "/Users/plainid/PycharmProjects/LDIF_Generator/venv/lib/python3.7/site-packages/ldif.py", line 139, in _unparseAttrTypeandValue
    if self._needs_base64_encoding(attr_type,attr_value):
  File "/Users/plainid/PycharmProjects/LDIF_Generator/venv/lib/python3.7/site-packages/ldif.py", line 128, in _needs_base64_encoding
    not safe_string_re.search(attr_value) is None
TypeError: cannot use a bytes pattern on a string-like object

Steps to reproduce: 
Enter example from documentation in python console

Fix is simple, mark the entry attribute strings as bytestrings:
entry={"objectClass":[b"top",b"person"],"cn":[b"Michael Stroeder"],"sn":[b"Stroeder"]}

Operating system: 
Mac OS
Python version: 
3.7
python-ldap version: 
3.2.0

atterdag:
Its not just python 3.7, I have the exact same issue with Python 3.6.8 on Ubuntu 18.04

atterdag:
Its kinda easy to solve in ldif.py. Of course this is a bit of a hack, but it works:

```
diff --git a/Lib/ldif.py b/Lib/ldif.py
index a26c8ac..2d12b10 100644
[...]

PR #317, opened 2020-01-29
Doc: Use bytes for attribute values in example

encukou:
Fixes: https://github.com/python-ldap/python-ldap/issues/306


encukou:
No review for 3 months :(
This is not a code change, so I'll merge without review.

PR #323, opened 2020-02-14
create BYTES_MODE_MODERN and BYTES_MODE_LEGACY constants

graingert:
I have terrible trouble remembering if `bytes_mode=True` or `bytes_mode=False` is the correct newer modern flag

### DIFF
--- a/Doc/bytes_mode.rst
+++ b/Doc/bytes_mode.rst
@@ -48,14 +48,15 @@ In Python 3, text values are represented as ``str``, the Unicode text type.
 In Python 2, the behavior of python-ldap 3.0 is influenced by a ``bytes_mode``
 argument to :func:`ldap.initialize`:
 
-``bytes_mode=True`` (backwards compatible):
+``bytes_mode=ldap.BYTES_MODE_LEGACY`` (backwards compatible):
     Text values are represented as bytes (``str``) encoded using UTF-8.
 
-``bytes_mode=False`` (future compatible):
+``bytes_mode=ldap.BYTES_MODE_MODERN`` (future compatible):
     Text values are represented as ``unicode``.
 
-If not given explicitly, python-ldap will default to ``bytes_mode=True``,
-but if an ``unicode`` value supplied to it, it will warn and use that value.
+If not given explicitly, python-ldap will default to
+``bytes_mode=ldap.BYTES_MODE_LGACY``, but if an ``unicode`` value is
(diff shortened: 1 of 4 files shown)


tiran:
bytes_mode only applies to Python 2 and Python 2 has reached its EOL. I'm not sure it makes sense to add a new feature for Python 2.

IIRC Petr plans to remove Python 2 support rather sooner than later. I guess there will be one more release with Python 2 support and latest fixes. After that we'll remove Python 2 code.

Doc/bytes_mode.rst
```diff
@@ -48,14 +48,15 @@ In Python 3, text values are represented as ``str``, the Unicode text type.
...
-``bytes_mode=True`` (backwards compatible):
+``bytes_mode=ldap.BYTES_MODE_LEGACY`` (backwards compatible):
     Text values are represented as bytes (``str``) encoded using UTF-8.
 
-``bytes_mode=False`` (future compatible):
+``bytes_mode=ldap.BYTES_MODE_MODERN`` (future compatible):
     Text values are represented as ``unicode``.
 
-If not given explicitly, python-ldap will default to ``bytes_mode=True``,
-but if an ``unicode`` value supplied to it, it will warn and use that value.
```
graingert: but if an ``unicode`` value *is* supplied to it, it will warn and use that value.

graingert:
@tiran sure but I'm going to be on python2 for a little while while porting it to py3

encukou:
This is bad timing for such an improvement. These constants can only be added to the *next* version of python-ldap, when it comes out. The number of people whom this will help is shrinking. And there is value in keeping the documentation unchanged for the benefit of all who have already read it.

As proposed, the documentation doesn't mention `bytes_mode=True` and `bytes_mode=False`, which are used in existing code. And mentioning them both would probably be too complex.

> I have terrible trouble remembering if `bytes_mode=True` or `bytes_mode=False` is the correct newer modern flag

Sorry for that. In retrospect, you're right, but I really think it's not worth the trouble of adding and documenting new constants.
As a workaround, add a comment to your code.

Does that sound fair?

graingert:
> Does that sound fair?

sure

Will you be making a version without the py2 workarounds?

encukou:
I'm in no rush to remove working code.
But if py2-only code makes some feature or bugfix harder to implement, I'll likely do a final release with Python 2 support, and drop Python 2 for the version after that.

PR #325, opened 2020-02-18
fix grammar in docs

graingert:


Doc/bytes_mode.rst
```diff
@@ -55,7 +55,7 @@ argument to :func:`ldap.initialize`:
     Text values are represented as ``unicode``.
 
 If not given explicitly, python-ldap will default to ``bytes_mode=True``,
-but if an ``unicode`` value supplied to it, it will warn and use that value.
+but if an ``unicode`` value is supplied to it, it will warn and use that value.
```
graingert:
```suggestion
but if a ``unicode`` value is supplied to it, it will warn and use that value.
```

graingert:
https://www.englishclub.com/pronunciation/a-an.htm

"ay you-ni-code"

encukou:
Thanks for your attention to detail!

graingert:
@encukou I've just read that line a few dozen times because I can't remember it :p

PR #330, opened 2020-05-06
Doc: Remove duplicate entry for NO_OBJECT_CLASS_MODS

encukou:
This fixes the build with newer versions of Sphinx.

(The entry is duplicated on line 444.)


encukou:
I'll merge this non-code change **without review** to unblock tests for other PRs.

Issue #326, opened 2020-04-08
Tox does not test the installed extension module


hroncok:
When running tests with tox, this happens:

 1. tox creates an sdist of current state of python-ldap in the working directory
 2. tox install (dependencies and) that sdist into the virtualenv, buulding the extension module for that Python version
 3. tox calls `setup.py test` and that builds the extension module again, testing the module from local build dir instead of the installed one

```
$ tox -e py38
GLOB sdist-make: .../python-ldap/setup.py
py38 inst-nodeps: .../python-ldap/.tox/.tmp/package/1/python-ldap-3.2.0.zip
py38 installed: coverage==5.0.4,pyasn1==0.4.8,pyasn1-modules==0.2.8,python-ldap==3.2.0
...
py38 run-test: commands[0] | .../python-ldap/.tox/py38/bin/python -bb -Werror '-Wignore:the imp module is deprecated:DeprecationWarning' '-Wignore:the imp module is deprecated:PendingDeprecationWarning' '-Wignore:Using or importing the ABCs from '"'"'collections'"'"' instead of from '"'"'collections.abc'"'"' is deprecated, and in 3.8 it will stop working:DeprecationWarning' -m coverage run --parallel setup.py clean --all test
running clean
'build/lib.linux-x86_64-3.8' does not exist -- can't clean it
'build/bdist.linux-x86_64' does not exist -- can't clean it
'build/scripts-3.8' does not exist -- can't clean it
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
running egg_info
writing Lib/python_ldap.egg-info/PKG-INFO
writing dependency_links to Lib/python_ldap.egg-info/dependency_links.txt
writing requirements to Lib/python_ldap.egg-info/requires.txt
writing top-level names to Lib/python_ldap.egg-info/top_level.txt
reading manifest file 'Lib/python_ldap.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
no previously-included directories found matching 'Doc/.build'
writing manifest file 'Lib/python_ldap.egg-info/SOURCES.txt'
running build_ext
building '_ldap' extension
creating build
creating build/temp.linux-x86_64-3.8
creating build/temp.linux-x86_64-3.8/Modules
[...]

PR #335, opened 2020-05-27
Tox: replace setup.py test with unittest discover

tiran:
setup.py test is deprecated and causes issues with testing our code.
There are cases where the wrong shared library is picked up.

The ``unittest discover`` approach avoids building the extension twice.

python-ldap should move to pytest eventually.

Fixes: https://github.com/python-ldap/python-ldap/issues/326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,10 +68,6 @@ matrix:
       env: TOXENV=doc
   allow_failures:
      - env:
-        - TOXENV=py39
-        - CFLAGS_std="-std=c99"
-        - WITH_GCOV=1
-     - env:
         - TOXENV=pypy
 
 env:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,10 @@
 [run]
 branch = True
 source =
-    Lib/
-omit =
-    Lib/slapdtest.py
+   ldap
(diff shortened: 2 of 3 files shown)


tiran:
The changeset breaks coverage and codecov. Back to the drawing board!

.travis.yml
```diff
@@ -88,7 +84,9 @@ env:
 
 install:
   - pip install "pip>=7.1.0"
-  - pip install tox-travis tox codecov coverage
+  - pip install tox-travis tox codecov
+  # See https://github.com/nedbat/coveragepy/issues/915
+  - pip install "coverage<5.0"
```
encukou, line 89: IMO this one can be reverted now, a fixed coverage 5.1 is out.
```suggestion
  - pip install tox-travis tox codecov coverage
```

tiran:
Let's try coverage 5.1

encukou:
For some reason, Travis CI is not reporting results. But the build for 68bfc62d20ea7483b645965c71db95d2e6d6365f passed: https://travis-ci.org/github/python-ldap/python-ldap/builds/695111114

hroncok:
Thanks!

PR #337, opened 2020-05-29
Allow LDAP connection from file descriptor 

tiran:
``ldap.initialize()`` now takes an optional fileno argument to create an
LDAP connection from a connected socket.

set_option() and get_option() now verify that LDAPObject is valid. This
fixes an assertion error and possible segfault after unbind_ext().

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -29,7 +29,7 @@ Functions
 
 This module defines the following functions:
 
-.. py:function:: initialize(uri [, trace_level=0 [, trace_file=sys.stdout [, trace_stack_limit=None, [bytes_mode=None, [bytes_strictness=None]]]]]) -> LDAPObject object
+.. py:function:: initialize(uri [, trace_level=0 [, trace_file=sys.stdout [, trace_stack_limit=None, [bytes_mode=None, [bytes_strictness=None, [fileno=None]]]]]]) -> LDAPObject object
 
    Initializes a new connection object for accessing the given LDAP server,
    and return an :class:`~ldap.ldapobject.LDAPObject` used to perform operations
@@ -39,6 +39,16 @@ This module defines the following functions:
    containing only the schema, the host, and the port fields. Note that
    when using multiple URIs you cannot determine to which URI your client
    gets connected.
+
+   If *fileno* parameter is given then the file descriptor will be used to
+   connect to an LDAP server. The *fileno* must either be a socket file
+   descriptor as :class:`int` or a file-like object with a *fileno()* method
+   that returns a socket file descriptor. The socket file descriptor must
+   already be connected. :class:`~ldap.ldapobject.LDAPObject` does not take
+   ownership of the file descriptor. It must be kept open during operations
+   and explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
+   unbound. The internal connection type is determined from the URI, ``TCP``
+   for ``ldap://`` / ``ldaps://``, ``IPC`` (``AF_UNIX``) for ``ldapi://``.
 
    Note that internally the OpenLDAP function
    `ldap_initialize(3) <https://www.openldap.org/software/man.cgi?query=ldap_init&sektion=3>`_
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -9,7 +9,8 @@ envlist = py27,py34,py35,py36,py37,py38,py39,{py2,py3}-nosasltls,doc,py3-trace,c
 minver = 1.8
 
 [testenv]
-deps = coverage
+# coverag 5.0.2 is failing with "sqlite3.OperationalError: disk I/O error"
+deps = coverage < 5.0
```
encukou, line 13: Can this be removed now? Would `coverage != 5.0` have been better?

Doc/reference/ldap.rst
```diff
@@ -40,6 +40,14 @@ This module defines the following functions:
...
    gets connected.
 
+   If *fileno* parameter is given the socket file descriptor will be used
+   to connect to an LDAP server. The socket file descriptor must already be
+   connected. :class:`~ldap.ldapobject.LDAPObject` does not take ownership
+   of the file descriptor. It must be kept open during operations and
+   explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
+   unbound. The internal connection type is determined from the URI, ``TCP``
+   for ``ldap://`` / ``ldaps://``, ``IPC`` for ``ldapi://``.
+
```
encukou, line 50: *fileno* can either be an integer or a file-like object with a `fileno()` method.

encukou, line 50: Would it not be better to allow closing the descriptor automatically when the object is unbound?

Modules/functions.c
```diff
@@ -30,6 +30,75 @@ l_ldap_initialize(PyObject *unused, PyObject *args)
     return (PyObject *)newLDAPObject(ld);
 }
 
+#ifdef HAVE_LDAP_INIT_FD
+
+/* initialize_fd(fileno, url)
+ *
+ * ldap_init_fd() is not a private API but it's not in a public header either
+ * SSSD has been using the function for a while, so it's probably OK.
```
encukou, line 38: Are there any plans to make it public?

tox.ini
tiran: Last time I tried the whole 5.x series of coverage was broken. There is some issue with sqlite.

Doc/reference/ldap.rst
tiran: In my experience it's less flexible and causes more trouble. You can't even use ``with socket.socket()`` because socket.close() will fail with an exception when the FD has been closed outside the socket code.

Modules/functions.c
tiran: I don't know why it hasn't landed in upstream headers. The API has been around for a while and it's stable. SSSD uses it heavily.

tox.ini
tiran: Let's see how the other PR ends up.

Doc/reference/ldap.rst
tiran: I have updated the documentation.

Modules/functions.c
encukou: Is there an RFE open?
I'm worried by the extern declaration: OpenLDAP is free to add a `ldap_init_fd` function to the public headers, and even make the declaration a bit different. (Would e.g. `unsigned int proto` be enough to cause python-ldap to fail to build?).

tiran: It's been document with the exact same function signature for 13 years, https://github.com/openldap/openldap/commit/282a3c537a9ab2f805a2f428606a3da3fa1705d5 . It's unlikely that the function will disappear or change signature.

encukou: OK.
Could you make sure OpenLDAP devs know, though?

quanah: > 
> 
> OK.
> Could you make sure OpenLDAP devs know, though?

The statement was false.  It is provided in the public openldap.h header file.

tiran: ... which is not packaged by distributions like Debian/Ubuntu or Fedora/RHEL.

quanah: > 
> 
> ... which is not packaged by most distributions like Debian/Ubuntu or Fedora/RHEL.

Then you need to talk to the distributions, not the OpenLDAP developers.

PR #343, opened 2020-06-05
Do not ignore warnings in tox config

encukou:
It looks like all the warnings were caused by setuptools.
Since we don't use setuptools in tests any more, we can remove the warnings.


tiran:
Ah, setuptools is the Wurst!

PR #345, opened 2020-06-05
Prepare release 3.3.0

encukou:


tiran:
There have been a bunch of changes to C code. I'd like to run the FreeIPA test suite with a pre-build of python-ldap first.

tiran:
PS: I also would like to deprecate Python 2.7, 3.4, and 3.5, and remove support in the version after 3.3.0.

encukou:
> There have been a bunch of changes to C code. I'd like to run the FreeIPA test suite with a pre-build of python-ldap first.

Could you also run the [additional test](https://www.python-ldap.org/en/python-ldap-3.2.0/contributing.html#additional-tests-and-scripts)? Some of them might require your knowledge.

> PS: I also would like to deprecate Python 2.7, 3.4, and 3.5, and remove support in the version after 3.3.0.

I'm planning to mention that in the release announcement, rather than in CHANGES.

tiran:
- FreeIPA's test suite is passing with master tip. I had to remove coverage in order to build test RPMs, though.
- valgrind doesn't show any considerable memory leaks.
- scan-build does not work for me. Looks like clang doesn't like some of the compiler arguments on Fedora: ``clang-10: error: unknown argument: '-fstack-clash-protection'``

encukou:
Oh, I thought you need a release to run the FreeIPA test suite!
Since that's not the case, I can skip the beta and release 3.3.0 immediately.

tiran:
I thought so, too. But I figured out how to hack a private build and submit it to FreeIPA's internal CI.

PR #350, opened 2020-06-07
Add high level LDAPObject.set_tls_options()

tiran:
The new high level function ``set_tls_options`` deals with most common
quirks and issues when setting TLS/SSL related options.


Tests/t_ldapobject.py
```diff
@@ -371,6 +371,62 @@ def test_multiple_starttls(self):
...
+    def test_set_tls_options_ldap(self):
+        # just any directory will do
+        certdir = os.path.dirname(__file__)
+        conn = self.ldap_object_class(self.server.ldap_uri)
+        conn.set_tls_options(
+            cacertfile=self.server.cafile,
+            # just any directory
+            cacertdir=certdir,
+            require_cert=ldap.OPT_X_TLS_DEMAND,
+            protocol_min=0x303,
```
droideck, line 387: It is a very minor nitpick.
I understand that it is common knowledge but can we mention here in a comment that 0x303 is TLS 1.2? For friendliness and explicitness:)

Lib/ldap/ldapobject.py
```diff
@@ -697,6 +698,127 @@ def set_option(self,option,invalue):
...
+      raise ValueError("libldap does not have TLS support")
+    # OpenSSL and GnuTLS support these options
+    tls_pkg = self.get_option(ldap.OPT_X_TLS_PACKAGE)
+    if tls_pkg not in {"OpenSSL", "GnuTLS"}:
+      raise ValueError("Unsupport TLS package '{}'.".format(tls_pkg))
+    # block ldapi ('in' because libldap supports multiple URIs)
+    if "ldapi://" in self._uri:
+      raise ValueError("IPC (ldapi) does not support TLS.")
+    if self._ldap_call(self._l.tls_inplace):
+      raise ValueError("TLS connection already established")
```
droideck, line 733: If I understand correctly the exception types... I think this one will fit a bit better - `EnvironmentError` (instead of `ValueError`).

Tests/t_ldapobject.py
```diff
@@ -371,6 +371,62 @@ def test_multiple_starttls(self):
             l.simple_bind_s(self.server.root_dn, self.server.root_pw)
             self.assertEqual(l.whoami_s(), 'dn:' + self.server.root_dn)
 
+    def assert_option_equal(self, conn, option, value):
+        self.assertEqual(conn.get_option(option), value)
+
+    @requires_tls()
+    def test_set_tls_options_ldap(self):
```
droideck, line 378: Looks like the coverage report has failed.
I think it's important to test all the parts before merging... 

If you haven't started on the test suite expansion yet, I can take a look later and add a few tests so it will cover the rest of the code. :)

encukou:
Speaking of friendliness, the added docs use a lot of acronyms that are very hard to search for if you don't already know what they mean.

encukou:
@quanah [mentioned](https://github.com/python-ldap/python-ldap/pull/431#issuecomment-921909308):
> Issue #350 still incorrectly states that you can't do start_tls over ldapi

I don't know much about this topic, but I'll note that `set_tls_options` doesn't need to be perfect when it's first merged. Everything it does can be done with lower-level API already.

PR #351, opened 2020-06-08
Syncrepl fails to parse SyncInfoMessage when the message is a syncIdSet

Firstyear:
Previously, the SyncInfoMessage parser used a for loop to test for the
presence of each named choice with with the structure. However, because
refreshDelete and refreshPresent both are able to be fully resolved
as defaults, and due to how pyasn1 accesses named types, it was not
checking the choice tag, and would return a fully populated refreshDelete
to the caller instead.

This fixes the parser to always return the inner component, and retrieves
the name based on the current choice of the tagged item.

Author: William Brown <william@blackhats.net.au>

### DIFF
--- a/Tests/t_ldap_syncrepl.py
+++ b/Tests/t_ldap_syncrepl.py
@@ -10,6 +10,7 @@ import os
 import shelve
 import sys
 import unittest
+import binascii
 
 if sys.version_info[0] <= 2:
     PY2 = True
@@ -21,7 +22,7 @@ os.environ['LDAPNOINIT'] = '1'
 
 import ldap
 from ldap.ldapobject import SimpleLDAPObject
-from ldap.syncrepl import SyncreplConsumer
+from ldap.syncrepl import SyncreplConsumer, SyncInfoMessage
 
 from slapdtest import SlapdObject, SlapdTestCase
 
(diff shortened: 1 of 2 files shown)


tiran:
Thanks for the fix, William.

Do you have a reproducer bug for the test suite?

Firstyear:
Added a test case for you.

Tests/t_ldap_syncrepl.py
```diff
@@ -445,6 +446,40 @@ def setUp(self):
...
+        3d6469726563746f7279206d616e616765723a64633d6578616d706c652c6463
+        3d636f6d3a286f626a656374436c6173733d2a2923330101ff311204108dc446
+        01a93611ea8aaff248c5fa5780
+        """.replace(' ', '').replace('\n', '')
+
+        msgraw = binascii.unhexlify(msg)
+        sim = SyncInfoMessage(msgraw)
+        self.assertEqual(sim.refreshDelete, None)
+        self.assertEqual(sim.refreshPresent, None)
+        self.assertNotEqual(sim.syncIdSet, None)
```
tiran, line 481:
```suggestion
        self.assertEqual(sim.newcookie, None)
        self.assertEqual(sim.syncIdSet,
            {
                'cookie': 'ldapkdc.example.com:38901#cn=directory manager:dc=example,dc=com:(objectClass=*)#3',
                'syncUUIDs': ['8dc44601-a936-11ea-8aaf-f248c5fa5780'],
                'refreshDeletes': True
            }
        )
```

Firstyear:
Thanks mate, appreciate the suggestion! I was a bit focused on the logic not the mesagge decode, but more coverage helps and helps guarantee correctness. Thanks!

tiran:
@encukou This fix is a candidate for 3.3.0 release, too.

Issue #353, opened 2020-06-09
python-ldap declares ldap_init_fd


encukou:
Currently, python-ldap carries a declaration of OpenLDAP's `ldap_init_fd`:
https://github.com/python-ldap/python-ldap/blob/c00694ee4b9b6d51bfbf75700adaff40b354e37c/Modules/functions.c#L47-L48

This function is [documented](https://github.com/openldap/openldap/commit/282a3c537a9ab2f805a2f428606a3da3fa1705d5) and declared in in `openldap.h`, but apparently [distros don't ship `openldap.h`](https://github.com/python-ldap/python-ldap/pull/337#discussion_r436125389).

Unless there is some complication that I don't see, we should not declare this function. Distros should either ship `openldap.h`, or they could patch `python-ldap` to provide the declaration.

encukou:
Apparently, `<openldap.h>` was added in OpenLDAP 2.4.48, only about 9 months ago: https://bugs.openldap.org/show_bug.cgi?id=8671

PR #354, opened 2020-06-09
Use openldap.h on OpenLDAP 2.4.48

tiran:
OpenLDAP 2.4.48 added openldap.h, which defines ldap_init_fd. Use the
header file with 2.4.48 and define the function for older versions.

The patch also cleans up #include. All OpenLDAP includes are now in
common.h and use global includes ``#include <ldap.h>`` instead of local
includes ``#include "ldap.h"``.

Fixes: https://github.com/python-ldap/python-ldap/issues/353
See: https://bugs.openldap.org/show_bug.cgi?id=8671

### DIFF
--- a/Modules/common.h
+++ b/Modules/common.h
@@ -25,9 +25,6 @@
    * see https://bugs.openldap.org/show_bug.cgi?id=8671
    */
 #include <openldap.h>
-#ifndef HAVE_LDAP_INIT_FD
-#define HAVE_LDAP_INIT_FD
-#endif
 #else
   /* ldap_init_fd() has been around for a very long time
    * SSSD has been defining the function for a while, so it's probably OK.
--- a/Modules/common.h
+++ b/Modules/common.h
@@ -12,6 +12,33 @@
 #include "config.h"
 #endif
 
+#include <lber.h>
+#include <ldap.h>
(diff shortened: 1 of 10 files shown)


Modules/common.h
```diff
@@ -12,6 +12,33 @@
...
+#endif
+#else
+  /* ldap_init_fd() has been around for a very long time
+   * SSSD has been defining the function for a while, so it's probably OK.
+   */
+#define LDAP_PROTO_TCP 1
+#define LDAP_PROTO_UDP 2
+#define LDAP_PROTO_IPC 3
+extern int ldap_init_fd(ber_socket_t fd, int proto, LDAP_CONST char *url,
+                        LDAP **ldp);
```
encukou, line 39: AFAICS, this declares `ldap_init_fd` but doesn't define `HAVE_LDAP_INIT_FD`, so the declaration is unused. What am I missing? What should `HAVE_LDAP_INIT_FD` be used for?

tiran: I added ``HAVE_LDAP_INIT_FD`` so users can disable the feature in case their libldap does not have the function. Since it's a public function now, has been around for a long time, and python-ldap requires OpenLDAP 2.4 any way, I'm going to remove the define again.

PR #355, opened 2020-06-18
Document the LDAPError fields

encukou:
Looking at the list of changfes, I realized the `LDAPError` info dict if poorly documented. This should help.

Also, this corrects the note on `COMPARE_FALSE` and `COMPARE_TRUE`, which was contradicting the `'msgid'` doc.

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -462,26 +462,23 @@ The module defines the following exceptions:
    are instead turned into exceptions, raised as soon an the error condition
    is detected.
 
-   The exceptions are accompanied by a dictionary possibly
-   containing an string value for the key :py:const:`desc`
-   (giving an English description of the error class)
-   and/or a string value for the key :py:const:`info`
-   (giving a string containing more information that the server may have sent).
+   The exceptions are accompanied by a dictionary with additional information.
+   The following fields are set in the dictionary if the relevant information
+   is available:
 
-   A third possible field of this dictionary is :py:const:`matched` and
-   is set to a truncated form of the name provided or alias dereferenced
-   for the lowest entry (object or alias) that was matched.
-
-   The field :py:const:`msgid` is set in the dictionary where the
-   exception can be associated with an asynchronous request.
-   This can be used in asynchronous code where :py:meth:`result()` raises the
-   result of an operation as an exception. For example, this is the case for
-   :py:meth:`compare()`, always raises the boolean result as an exception
-   (:py:exc:`COMPARE_TRUE` or :py:exc:`COMPARE_FALSE`).
-
-   Most exceptions from protocol results also carry the :py:attr:`errnum`
-   attribute.
-
+   * ``'desc'``: string giving an English description of the error class.
+   * ``'info'``: string containing more information that the server may
+     have sent.
+   * ``'errnum'``: numeric error code.
+   * ``'matched'``: truncated form of the name provided or alias.
+     dereferenced for the lowest entry (object or alias) that was matched.
+   * ``'msgid'``: ID of the matching asynchronous request.
+     This can be used in asynchronous code where :py:meth:`result()` raises the
+     result of an operation as an exception. For example, this is the case for
+     :py:meth:`~LDAPObject.compare()`, always raises the boolean result as an
+     exception (:py:exc:`COMPARE_TRUE` or :py:exc:`COMPARE_FALSE`).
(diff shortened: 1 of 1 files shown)


Doc/reference/ldap.rst
```diff
@@ -462,26 +462,23 @@ The module defines the following exceptions:
...
-   (:py:exc:`COMPARE_TRUE` or :py:exc:`COMPARE_FALSE`).
-
-   Most exceptions from protocol results also carry the :py:attr:`errnum`
-   attribute.
-
+   The exceptions are accompanied by a dictionary with additional information.
+   The following fields are set in the dictionary if the relevant information
+   is available:
+
+   * ``'desc'``: string giving an English description of the error class.
```
tiran, line 469: It might be worth mentioning that the message is coming from libldap's ``ldap_err2string``.

The strings are not necessarily in English. libldap supports i18n through gettext and the error messages are marked for translation with ``N_("message")``.

Doc/reference/ldap.rst
```diff
@@ -462,26 +462,23 @@ The module defines the following exceptions:
...
+   * ``'info'``: string containing more information that the server may
```
tiran, line 470: It might be worse mentioning that the string is server specific. OpenLDAP server may send different info messages than AD or 389-DS.

Doc/reference/ldap.rst
```diff
@@ -462,26 +462,23 @@ The module defines the following exceptions:
...
-   result of an operation as an exception. For example, this is the case for
-   :py:meth:`compare()`, always raises the boolean result as an exception
-   (:py:exc:`COMPARE_TRUE` or :py:exc:`COMPARE_FALSE`).
-
-   Most exceptions from protocol results also carry the :py:attr:`errnum`
-   attribute.
-
+   The exceptions are accompanied by a dictionary with additional information.
+   The following fields are set in the dictionary if the relevant information
+   is available:
```
tiran, line 467: Could you please rephrase the sentence so it's really obvious that all members are optional and that python-ldap may add more members in the future. Or are any of the values guaranteed to be present?

Doc/reference/ldap.rst
```diff
@@ -462,26 +462,23 @@ The module defines the following exceptions:
...
+
+   * ``'desc'``: string giving an English description of the error class.
+   * ``'info'``: string containing more information that the server may
+     have sent.
+   * ``'errnum'``: numeric error code.
```
tiran, line 472: I don't see ``errnum`` in ``LDAPraise_for_message``. There are ``errno`` and ``result``, though.

Doc/reference/ldap.rst
encukou: I'll call them all optional. If any are guaranteed, they should be made into attributes.

Doc/reference/ldap.rst
encukou: Hm, that's right. I copied blindly :(
`result` comes from `LDAP_OPT_ERROR_NUMBER`, while `errno` is the C global.
I don't understand the difference too deeply, do you have an idea about how to best document these?

encukou: If not, I'll just say where they come from.

Doc/reference/ldap.rst
```diff
@@ -462,26 +462,25 @@ The module defines the following exceptions:
...
+   The exceptions are accompanied by a dictionary with additional information.
+   All fields are optional and more fields may be added in the future.
+   Currently, ``python-ldap`` may set the following fields:
+
+   * ``'desc'``: string giving a description of the error class, as provided
+     by OpenLDAP's ``ldap_err2string``.
+   * ``'info'``: string containing more information that the server may
+     have sent. The value is server-specific: for example, the OpenLDAP server
+     may send different info messages than Active Directory or 389-DS.
+   * ``'errno'``: numeric error code.
```
tiran, line 474: If I understand the code correctly than ``LDAPraise_for_message`` sets **two** different numeric error codes

* ``'errno'``: numeric errno from system call, libc, or C function calls (see :mod:`errno`)
* ``'result'``: numeric LDAP protocol error number or API error number

``result`` is set ``errnum`` converted to Python int, ``desc`` is basically ``ldap_err2string(errnum)``. ``errno`` and ``errno`` are different things.

encukou:
Alright! Hopefully this wording explains it well.

encukou:
> Much better than before

Mostly thanks to your comments! :)

PR #358, opened 2020-06-19
Remove Python 2 support

tiran:
python-ldap 3.4 will require Python 3.6 or newer.


encukou:
I added my suggestions in the form of commits; I hope you don't mind. Do they look OK?

tiran:
+1
Thanks!

Issue #359, opened 2020-06-20
ImportError on MacOS 10.15.5


OutOfFocus4:
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

## Issue description:

After installing python-ldap 3.3.0 on my iMac, it raises an `ImportError` whenever I try to import it.

## Steps to reproduce:
```sh
cd /tmp
python3 -m venv test
source test/bin/activate
pip install -U pip setuptools
pip install python-ldap
python -c 'import ldap'
```

Output:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/private/tmp/test/lib/python3.7/site-packages/ldap/__init__.py", line 34, in <module>
    import _ldap
ImportError: dlopen(/private/tmp/test/lib/python3.7/site-packages/_ldap.cpython-37m-darwin.so, 2): Symbol not found: _ldap_init_fd
  Referenced from: /private/tmp/test/lib/python3.7/site-packages/_ldap.cpython-37m-darwin.so
  Expected in: flat namespace
 in /private/tmp/test/lib/python3.7/site-packages/_ldap.cpython-37m-darwin.so
```

tiran:
``Symbol not found: _ldap_init_fd`` points to a new feature that was introduced a couple of weeks ago. What's your OpenLDAP libdap version? Are you using libldap from homebrew or from macOS?

xroot88:
Hi tiran, it says in the original comment: Python version: CPython 3.7.7 (installed with Homebrew)

OutOfFocus4:
I am using the MacOS system `libldap`. According to the man page, it is OpenLDAP 2.4.28.

tiran:
Thanks for the information, PR #360 should fix the problem.

@encukou can you do a 3.3.1 release on Monday?

flymg:
until merge, using `python-ldap==3.2.0` is the workaround

tiran:
I recommend to use ``python-ldap!=3.3.0`` instead. The requirements blocks 3.3.0 and install 3.2.0 for now. It will pick up 3.3.1 as soon as the fix is released.

encukou:
This should be fixed in the fresh python-ldap 3.3.1.
Thank you for the wait!

mvreijn:
Even though this should be fixed, I run into the same error on my MBP with Big Sur and homebrew LDAP libs (2.5.8): 

```
[...]

PR #360, opened 2020-06-20
Fix macOS SDK builds without ldap_init_fd

tiran:
macOS system libldap 2.4.28 does not have ldap_init_fd symbol. Disable
initialize_fd when Apple libldap 2.4.28 is detected.

Also run some macOS tests on Travis CI. Since the SDK does not ship
slapd, testing is rather limited.

Fixes: https://github.com/python-ldap/python-ldap/issues/359

### DIFF
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -49,7 +49,7 @@ This module defines the following functions:
    and explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
    unbound. The internal connection type is determined from the URI, ``TCP``
    for ``ldap://`` / ``ldaps://``, ``IPC`` (``AF_UNIX``) for ``ldapi://``.
-   The is not available on macOS when python-ldap is compiled with system
+   The parameter is not available on macOS when python-ldap is compiled with system
    libldap, see :py:const:`INIT_FD_AVAIL`.
 
    Note that internally the OpenLDAP function
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -49,6 +49,8 @@ This module defines the following functions:
    and explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
    unbound. The internal connection type is determined from the URI, ``TCP``
    for ``ldap://`` / ``ldaps://``, ``IPC`` (``AF_UNIX``) for ``ldapi://``.
+   The is not available on macOS when python-ldap is compiled with system
+   libldap, see :py:const:`INIT_FD_AVAIL`.
 
    Note that internally the OpenLDAP function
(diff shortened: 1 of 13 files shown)


encukou:
Could you also document when the functionality is available?

tiran:
> Could you also document when the functionality is available?

I would liek to but I don't know. OpenLDAP 2.4.28 has ``ldap_init_fd``. It was introduced around 2.4.0. I have no clue why the symbol is missing in the Apple builds. It could be patched out or hidden by a downstream patch.

encukou:
From the patch, I gather that on a Mac, you won't get fileno support even if you compile OpenLDAP 2.4.28 yourself.

tiran:
> From the patch, I gather that on a Mac, you won't get fileno support even if you compile OpenLDAP 2.4.28 yourself.

2.4.28 was released over 8 years ago. I don't we have to care about ancient versions except for Apple system builds.

I'm also open to better ways to detect system libldap on macOS. I don't have access to macOS to investigate the issue.

encukou:
I'm OK with the patch itself. I just think it needs a note in the documentation that says when the feature is available. From the current text, it seems that *fileno* is always usable.

Also, you've now rebased onto master, which I don't understand. Shouldn't this go into 3.2.1?

tiran:
> I'm OK with the patch itself. I just think it needs a note in the documentation that says when the feature is available. From the current text, it seems that _fileno_ is always usable.

I have updated documentation to mention ``INIT_FD_AVAIL`` and the macOS issue.

> Also, you've now rebased onto master, which I don't understand. Shouldn't this go into 3.3.1?

The fix should land in master and 3.3.1. New features land in master and then are back-ported to bug fix branches.

tiran:
Please merge #361 and #362 before this PR.

encukou:
> New features land in master and then are back-ported to bug fix branches.

That way makes it harder to merge. It is the CPython workflow, but this repo doesn't have Miss Islington like CPython.

tiran:
> That way makes it harder to merge. It is the CPython workflow, but this repo doesn't have Miss Islington like CPython.

I use the same workflow in all projects. Patches land in master first and then get backported to maintenance branches. The patch would have conflicted either way.

To make it easier for you I have created a backport branch, https://github.com/python-ldap/python-ldap/pull/363

Doc/reference/ldap.rst
```diff
@@ -49,6 +49,8 @@ This module defines the following functions:
    and explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
    unbound. The internal connection type is determined from the URI, ``TCP``
    for ``ldap://`` / ``ldaps://``, ``IPC`` (``AF_UNIX``) for ``ldapi://``.
+   The is not available on macOS when python-ldap is compiled with system
```
encukou, line 52:
```suggestion
   The parameter is not available on macOS when python-ldap is compiled with system
```

PR #363, opened 2020-06-29
[3.3] Fix macOS SDK builds without ldap_init_fd

tiran:
macOS system libldap 2.4.28 does not have ldap_init_fd symbol. Disable
initialize_fd when Apple libldap 2.4.28 is detected.

Also run some macOS tests on Travis CI. Since the SDK does not ship
slapd, testing is rather limited.
    
Fixes: https://github.com/python-ldap/python-ldap/issues/359


Doc/reference/ldap.rst
```diff
@@ -49,6 +49,8 @@ This module defines the following functions:
    and explicitly closed after the :class:`~ldap.ldapobject.LDAPObject` is
    unbound. The internal connection type is determined from the URI, ``TCP``
    for ``ldap://`` / ``ldaps://``, ``IPC`` (``AF_UNIX``) for ``ldapi://``.
+   The is not available on macOS when python-ldap is compiled with system
```
encukou, line 52:
```suggestion
   The parameter is not available on macOS when python-ldap is compiled with system
```

PR #364, opened 2020-06-29
Prepare auto-formatting with black

tiran:
Also removes autopep8 and prepares configuration for import sorter
isort.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,8 @@
+[tool.black]
+line-length = 88
+target-version = ['py36', 'py37', 'py38']
+
+[tool.isort]
+line_length=88
+known_first_party=['ldap', '_ldap', 'ldapurl', 'ldif', 'slapdtest']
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.black]
+line-length = 88
```
encukou, line 2: If this needs to be specified, why not stick to PEP 8?

tiran: 88 is the default line length for black. I'm mentioning 88 here to make the line length setting explicit.

encukou:
Overall, I must say I don't agree with the choices Black enforces. I won't enjoy reviewing these changes.

I would like to step down as the main maintainer of python-ldap before this goes in.

droideck:
Personally, I like the `black` choices. It is closer to `prettier` and I get used to it.
It is PEP8 compliant plus it has increased readability.
You have my ack.

PR #376, opened 2020-09-01
Mention pytest-ldap in the documentation

### LINKED ISSUE #372: Pytest fixtures
python-ldap version: 3.3.0

There is some [unittest testcases](https://www.python-ldap.org/en/python-ldap-3.3.0/reference/slapdtest.html) provided by python-ldap. It would be awesome if it could also provide some pytest fixtures.

I think there is not a lot of work, I can provide a patch. I just want to check if this patch is wanted, else I can create a `pytest-ldap` module.

What do you think?

azmeuk:
Fixes #372

### DIFF
--- a/Doc/reference/slapdtest.rst
+++ b/Doc/reference/slapdtest.rst
@@ -14,6 +14,8 @@ This module is pure Python and does not rely on any non-standard modules.
 Therefore it can be used stand-alone without the rest of the python-ldap
 package.
 
+For pytest fixtures, check `pytest-ldap`_.
+
 Functions
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


Doc/reference/slapdtest.rst
```diff
@@ -14,6 +14,8 @@ This module is pure Python and does not rely on any non-standard modules.
 Therefore it can be used stand-alone without the rest of the python-ldap
 package.
 
+For pytest fixtures, check `pytest-ldap`_.
```
encukou, line 17:
```suggestion
Test fixtures for the popular `pytest` framework are developed in an external project, `pytest-ldap`_.
```

encukou (approved):

encukou:
Thank you!
I reworded it a bit so it's more understandable to people who don't know pytest yet.

PR #378, opened 2020-09-01
tox positional arguments

azmeuk:
Tox can pass positional arguments to test runners. This allows for instance to debug a single test with `tox -e py38 -- -f -k test_foobar`.


encukou (approved):
Looks useful, thanks!

PR #382, opened 2020-10-01
SlapdObject directory based configuration method, and slapadd implementation

### LINKED ISSUE #373: Add a custom schema to SlapdObject with a slapadd method?
Operating system: Archlinux
Python version: 3.8.5
python-ldap version: 3.3.0

I wrote a custom schema and I would like to write some tests using this schema.

I can manage to do this by copying or linking my schema files in `SlapdObject.SCHEMA_DIR` before the process is launched. However this is not convenient at all (see next comment).

The `ldapadd` command cannot add schemas:

```python
RuntimeError: ['/usr/bin/ldapadd', '-H', 'ldapi://%2Fhome%2Feloi%2Fdev%2Fyaal%2Foidc-ldap-bridge%2Fpython-ldap-test-52629%2Fldapi', '-Y', 'EXTERNAL', '-Q'] process failed:
b'adding new entry "cn=oauth,cn=schema,cn=config"\n\n'
b'ldap_add: Insufficient access (50)\n'
```

However, the `slapadd` command could. I suggest to implement a `SlapdObject.slapadd` method, similar to `SlapdObject.ldapadd`. This would allow to call `slapadd -l myschema.ldif -n0` (`-n0` being an option, that could be used to add the schema in the `config` database). This way, one could add a schema to the `slapd` server test after it is started.

If `ldapadd` seems too much, maybe we can just find a mechanism to simply add custom schemas by configuring or passing parameters to `SlapdObject`?

If you are ok with this, I can provide a patch.
What do you think?

### LINKED ISSUE #377: SlapdObject configuration methods
python-ldap version: 3.3.0

SlapdObject creates a server configured by a configuration file, and this method is deprecated in favor of a configuration directory. Also it seems this prevents some actions to be doable with `slapadd`, that I would like to implement (see #373).

What would you think of making SlapdObject spawn a directory configured slapd server? This could be optional or by default.

I volunteer to provide a patch.

azmeuk:
Fixes #373 
Fixes #377

[This OpenLDAP documentation](https://www.openldap.org/doc/admin24/slapdconf2.html) tells:

> The older style slapd.conf(5) file is still supported, but its use is deprecated and support for it will be withdrawn in a future OpenLDAP release. Configuring slapd(8) via slapd.conf(5) is described in the next chapter.

This patch:
- makes slapd use the new directory based configuration method;
- implements the `SlapdObject.slapadd method`;
- tests the usage of `SlapdObject.slapadd method` by adding a custom schema;
- allow users to redefine `SlapdObject.openldap_schema_files` so they can load custom schemas at startup.

### DIFF
--- a/Doc/spelling_wordlist.txt
+++ b/Doc/spelling_wordlist.txt
@@ -129,6 +129,7 @@ serverctrls
 sessionSourceIp
 sessionSourceName
 sessionTrackingIdentifier
+slapadd
 sizelimit
 slapd
 stderr
--- a/Doc/spelling_wordlist.txt
+++ b/Doc/spelling_wordlist.txt
@@ -132,6 +132,7 @@ sessionTrackingIdentifier
 slapadd
 sizelimit
 slapd
+startup
 stderr
 stdout
 str
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -174,6 +174,9 @@ class SlapdObject(object):
     manager, the slapd server is shut down and the temporary data store is
     removed.
 
+    :param openldap_schema_files: A list of schema names or schema paths to
+        load at startup. By default this only contains `core`.
+
     .. versionchanged:: 3.1
 
         Added context manager functionality
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -23,34 +23,33 @@ import ldap
 
 HERE = os.path.abspath(os.path.dirname(__file__))
 
-# a template string for generating simple slapd.conf file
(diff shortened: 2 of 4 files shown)


azmeuk:
@encukou Can I do something to help you reviewing this patch? Should we summon other contributors?

droideck:
Looks good to me!
If @encukou, @tiran, or others don't any objections, I think, we can push it.

droideck (approved):
Looks good to me!
After @encukou, @tiran, or another contributor will take a look and give their review, we can push it.

encukou:
> Should we summon other contributors?

Yes. My role here is mostly administrative.
@tiran, if you don't have any objections, I'll merge.

PR #383, opened 2020-10-01
Python 3 syntax

azmeuk:
I applied the [pyupgrade](https://github.com/asottile/pyupgrade) tool to use a modern python syntax in the code.

```bash
find . -name "*.py" -exec pyupgrade --py36-plus "{}" +
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ PKG-INFO
 
 # generated in the sample workflow
 /__venv__/
+
+# test dirs
+python-ldap-test-*
--- a/Demo/Lib/ldap/async/deltree.py
+++ b/Demo/Lib/ldap/async/deltree.py
@@ -1,5 +1,3 @@
-from __future__ import print_function
-
 import ldap,ldap.async
 
 class DeleteLeafs(ldap.async.AsyncSearchHandler):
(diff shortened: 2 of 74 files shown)


encukou:
It looks like the tool changes `'%s' % ...` formatting to `'{}'.format(...)`, and `'{}'.format(...)` to `f'{...}'`, but doesn't do the two steps at once.
Could you try running it again, to hopefully get a more consistent result?

azmeuk:
The second run does not change anything :(

droideck (approved):
After looking for some time through the changes, I tend to think that the change is good as it is now.

I think it is okay to use either `str.format()`, either `f"str"`.
Mostly I prefer the second option but sometimes it'll be more readable to specify the first one, i.e.:
```
        return '<{}.{} instance at {}: {}>'.format(
            self.__class__.__module__,
            self.__class__.__name__,
            hex(id(self)),
            self._data
        )
```

So, I give my ack for the change as it's an obvious improvement and we should really get rid of the old string formatting '%'.

But as always, we need a second ack here. :) Could you please double-check me, guys?
@tiran @encukou

encukou (approved):
Looks good to me. Thank you!

Issue #387, opened 2020-11-06
cidict no longer defines a .copy() method


mistotebe:
Issue description:

At some point, cidict.copy() disappeared.

Steps to reproduce: 

```python3
from ldap.cidict import cidict
d = cidict()
d.copy()
```

Used to work, now `.copy()` is no longer a method.

Operating system: Linux
Python version: 3.8
python-ldap version: 3.3.1

encukou:
The `cidict` API was never really documented, but I think that for backwards compatibility reasons, `copy` should be part of it.

AFAIK that's the only API `dict` had in Python 2 but `MutableMapping` doesn't have. (We don't need to implement `__or__` or `__reversed__`, which `dict` got relatively recently.)

tiran:
``collections.UserDict`` implements a bunch of extra methods on top of MutableMapping. copy() is not provied by MutableMapping.

mistotebe:
Do you have a timeline when this might get released?

PR #388, opened 2020-11-06
Add missing cidict.copy() method

tiran:
The copy() method vanished when cidict was re-implemented on top of
MutableMapping abstract base class.

See: commit 2a2cef38feb7f259ea6b5735a6ce3e46812e217d
Fixes: https://github.com/python-ldap/python-ldap/issues/387


encukou:
Thank you!

Issue #203, opened 2018-04-03
Package slapdtest as a stand-alone library


encukou:
slapdtest is now positioned as an internal testing library, but it's useful elsewhere and other projects are starting to depend on it. It would be healthier for it to be released separately. Of course python-ldap's should still drive its development, but not exclusively.

Currently, come parts of it are genuinely python-ldap-specific, mainly the decorators (`skip_unless_ci`, `requires_tls`, …). python-ldap will need a new place for them.

Infrastructure-wise, slapdtest should be a new project under the python-ldap GitHub organization.

tiran:
Why do you consider it healthier to release slapdtest as a separate project? We can enforce stricter backwards compatibility either way. A separate library increases your workload. It also creates additional churn when we have to change both slapdtest and tests for a new feature or bug fix.

I see only one advantage: ``pip install python-ldap`` would create a slightly smaller installation because ``slapdtest`` and test helpers are no longer installed by default.

Some of the test decorators like ``requires_tls``, ``requires_sasl``, and ``requires_ldapi`` are useful for 3rd party projects, too. ``skip_unless_ci`` belongs into ``python-ldap``.

encukou:
> Why do you consider it healthier to release slapdtest as a separate project? We can enforce stricter backwards compatibility either way.

It's not about being stricter, but rather about providing different release cycles and guarantees.
Right now, for example, a project can't use one version of python-ldap internally but a newer version or slapdtest for its own test suite.
It's also about setting scope: "anything the python-ldap test suite happens to need" is not something I want pople to reuse.

> A separate library increases your workload.
> It also creates additional churn when we have to change both slapdtest and tests for a new feature or bug fix.

Valid concerns. Perhaps the libraries should be developed in the same repo, and just be released separately?

> I see only one advantage: pip install python-ldap would create a slightly smaller installation because slapdtest and test helpers are no longer installed by default.

> Some of the test decorators like requires_tls, requires_sasl, and requires_ldapi are useful for 3rd party projects, too. skip_unless_ci belongs into python-ldap

True, but we need to draw the line somewhere.
I think it should be at “spawning and controlling `slapd`”, but I can be persuaded otherwise.

tiran:
> True, but we need to draw the line somewhere.
> I think it should be at “spawning and controlling slapd”, but I can be persuaded otherwise.

[...]

PR #389, opened 2020-11-08
Make parameters configurable for SlapdObject

azmeuk:
A minor improvement for SlapdObject for cases we do want to specify:
- host 
- port

Also, those parameters can be passed to `SlapdObject.__init__` so we do not have to build a class that inherits from SlapObject just to configure this:
- log_level
- initial schemas
- database type
- suffix
- admin cn
- admin password

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -224,9 +224,17 @@ class SlapdObject(object):
     # create loggers once, multiple calls mess up refleak tests
     _log = None
 
-    def __init__(self, host=None, port=None, log_level=logging.WARN):
+    def __init__(self,
+                 host=None,
+                 port=None,
+                 log_level=logging.WARN,
+                 openldap_schema_files=None,
+                 ):
         if SlapdObject._log is None:
             SlapdObject._log = combined_logger('python-ldap-test', log_level=log_level)
+
+        if openldap_schema_files is not None:
+            self.openldap_schema_files = openldap_schema_files
 
         self._proc = None
         self._port = port or self._avail_tcp_port()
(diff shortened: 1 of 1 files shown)


Lib/slapdtest/_slapdtest.py
```diff
@@ -125,7 +125,7 @@ def _add_sbin(path):
 
 def combined_logger(
         log_name,
-        log_level=logging.WARN,
+        log_level,
```
encukou, line 128: Please keep the default value, so that the API doesn't change for other callers.

Lib/slapdtest/_slapdtest.py
```diff
@@ -212,16 +222,44 @@ class SlapdObject(object):
...
+                 log_level=logging.WARN,
+                 openldap_schema_files=None,
+                 database=None,
+                 suffix=None,
+                 root_cn=None,
+                 root_pw=None,
+                 datadir_prefix=None,
+                 ):
+        if SlapdObject._log is None:
+            SlapdObject._log = combined_logger('python-ldap-test', log_level=log_level)
```
encukou, line 239: This will change the logger for *all* SlapdObject instances.
Please use `self._log` instead, only set this if `log_level` is not using the default, and ideally change the logger name to reflect the log_level.

azmeuk:
I have fixed your suggestions.

droideck (approved):
Looks like a good clean up and improvement! Thank you!
As the CI tests are passing and the code seems reasonable - you have my ack.

azmeuk:
I included this patch on `python-slapd`.
https://github.com/python-ldap/python-ldap/issues/203#issuecomment-731778829

PR #390, opened 2020-11-09
SlapdObject increasing waiting time

azmeuk:
This gives a chance to SlapdObject to start faster. Notably it does not start by sleeping 1.5s anymore.


tiran:
The ``range()`` timeout loop is flawed because it may wait much longer than we want. A better approach is the set a deadline with a monotonic timer, do the work, check deadline and sleep in a while loop:

```
deadline = time.monotonic() + 10
while True:
    # do stuff
    if time.monotonic() >= deadline:
        break
    time.sleep(0.2)
```

azmeuk:
Thank you, this is indeed a much cleaner way. This is fixed.

droideck (approved):
Looks good to me!

azmeuk:
I just noticed that this patch had greatly diminished the `tox` execution time \o/

PR #391, opened 2020-11-09
Removed some python2 related code from SlapdObject

azmeuk:


droideck (commented):
As you are doing the cleanup around Python 2 area...
Could you please remove this part too?

https://github.com/python-ldap/python-ldap/blob/master/Modules/LDAPObject.c#L754

It is the only other place I found where Python 2 is mentioned.

PR #393, opened 2020-11-10
SlapdObject command methods return process data

azmeuk:
A simple refactoring, so the output of commands is usable.

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -518,7 +518,7 @@ class SlapdObject:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE
         )
-        self._log.debug('stdin_data=%r', stdin_data)
+        self._log.debug('stdin_data=%s', stdin_data.decode("utf-8") if stdin_data else stdin_data)
         if proc.stdout is not None:
             self._log.debug('stdout=%s', proc.stdout.decode("utf-8"))
         if proc.stderr is not None:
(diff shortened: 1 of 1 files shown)


droideck (dismissed):
A very small nitpick... 
Could you please use an imperative mood in the subject line? It makes the changelog much more readable.

Besides that, you have my ack!

azmeuk:
Something like "Make SlapdObject commands return their outputs" would fit better?

encukou:
Before the API is set in stone, are we sure it's the right one? Would it be better to return something like [`subprocess.CompletedProcess`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess)?

Could you also add the return value to the docstrings and documentation?

azmeuk:
@encukou I did not know about `CompletedProcess` but this seems the cleaner way to do this!
I have fixed the patch (sorry for the force-push).

Lib/slapdtest/_slapdtest.py
```diff
@@ -512,60 +512,69 @@ def _cli_popen(self, ldapcommand, extra_args=None, ldap_uri=None,
...
+        if proc.stdout is not None:
+            self._log.debug('stdout=%s', proc.stdout.decode("utf-8"))
+        if proc.stderr is not None:
+            self._log.debug('stderr=%s', proc.stderr.decode("utf-8"))
+        if proc.returncode != 0:
             raise RuntimeError(
-                '{!r} process failed:\n{!r}\n{!r}'.format(
-                    args, stdout_data, stderr_data
-                )
+                'Process failed: {!r}'.format(" ".join(args))
```
azmeuk, line 528: I removed `stdout` and `stderr` here as they have been logged a few lines ago, and large outputs as exceptions messages are totally unpracticable.

droideck (approved):
Looks clean and improved!
You have my ack.

azmeuk:
I included this patch on `python-slapd`.
https://github.com/python-ldap/python-ldap/issues/203#issuecomment-731778829

PR #395, opened 2020-11-14
Test with GHA

tiran:


encukou:
This looks reasonable. What's keeping it as draft?

encukou:
The checks didn't run here, but I assume that's due to some bootstrapping issue. They did pass for the same commit at https://github.com/tiran/python-ldap/pull/1
Since travis-ci.org is down, this is better than the status quo.

PR #397, opened 2020-11-18
ppolicy control documentation

azmeuk:
I just noticed that the `ppolicy` control did not appear in the documentation.

Attributes are auto-documented with [sphinx autoattribute](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute)

> For module data members and class attributes, documentation can either be put into a comment [...] or in **a docstring after the definition**.

### DIFF
--- a/Lib/ldap/controls/ppolicy.py
+++ b/Lib/ldap/controls/ppolicy.py
@@ -62,17 +62,31 @@ class PasswordPolicyResponseValue(univ.Sequence):
 
 
 class PasswordPolicyControl(ValueLessRequestControl,ResponseControl):
+  """
+  Indicates the errors and warnings about the password policy.
+
+  Attributes
+  ----------
+
+  timeBeforeExpiration : int
+      The time before the password expires.
+
+  graceAuthNsRemaining : int
+      The number of grace authentications remaining.
+
+  error: int
+      The password and authentication errors.
+  """
   controlType = '1.3.6.1.4.1.42.2.27.8.5.1'
 
   def __init__(self,criticality=False):
     self.criticality = criticality
-
-  def decodeControlValue(self,encodedControlValue):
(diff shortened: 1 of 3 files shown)


Lib/ldap/controls/ppolicy.py
```diff
@@ -62,17 +62,25 @@ class PasswordPolicyResponseValue(univ.Sequence):
...
+  """
   controlType = '1.3.6.1.4.1.42.2.27.8.5.1'
 
   def __init__(self,criticality=False):
     self.criticality = criticality
 
-  def decodeControlValue(self,encodedControlValue):
-    ppolicyValue,_ = decoder.decode(encodedControlValue,asn1Spec=PasswordPolicyResponseValue())
     self.timeBeforeExpiration = None
+    """The time before the password expires."""
```
tiran, line 74: This style of inline comments is uncommon and won't affect documentation. We don't use ``epydoc`` or ``napoleon`` markup to document attributes yet.

tiran (changes requested):

Lib/ldap/controls/ppolicy.py
azmeuk: I am not sure what you suggest. This is the type of inline comment sphinx uses in their samples in the link I provided. I checked the generated documentation, and it displays what I expected to.

tiran: I see. Sphinx seems to parse the code to find these extra lines. Inline strings are ignored by Python. I don't like them because they are extra noise that district from code. Instead I would prefer to use Napoleon, epydoc, or at least ``#:`` comments.

azmeuk: This should be better now.

droideck (approved):
The change looks good to me!
Good catch with the `decodeControlValue` method!

encukou:
Thank you!

PR #398, opened 2020-11-18
Make sure the documentation copyright year is always up-to-date

azmeuk:


Doc/conf.py
```diff
@@ -53,7 +54,7 @@
 
 # General substitutions.
 project = 'python-ldap'
-copyright = '2008-2017, python-ldap project team'
+copyright = '2008-{}, python-ldap project team'.format(datetime.datetime.now().year)
```
encukou, line 57: If we can't give the exact years when we made some copyrightable changes, I'd rather leave the year out entirely.
AFAIK, copyright is automatic; these notices are just a friendly reminders.
```suggestion
copyright = 'python-ldap project team'
```

azmeuk: Ok. I do not have a strong opinion about this. I think no date is better than an old date.

droideck (approved):
Looks good. Ack!

PR #410, opened 2021-05-06
Use slapd.Slapd instead of SlapdObject in unit tests

### LINKED ISSUE #203: Package slapdtest as a stand-alone library

azmeuk:
Fixes #203
All the tests run with [python-slapd](https://github.com/python-ldap/python-slapd) instead of `slapdtest.SlapdObject`.

### DIFF
--- a/Lib/slapdtest/_slapdtest.py
+++ b/Lib/slapdtest/_slapdtest.py
@@ -10,6 +10,7 @@ import time
 import subprocess
 import logging
 import atexit
+import slapd
 from logging.handlers import SysLogHandler
 import unittest
 from shutil import which
@@ -584,7 +585,7 @@ class SlapdTestCase(unittest.TestCase):
     test class which also clones or initializes a running slapd
     """
 
-    server_class = SlapdObject
+    server_class = slapd.Slapd
     server = None
     ldap_object_class = None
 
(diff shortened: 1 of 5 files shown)


tiran:
I feel bad for saying this and for all the work you have put into your PR ... but I'm strongly **against** this change. It would cause additional work and complicated for both my work as a maintainer of this package and even more as a downstream package maintainer for Fedora, CentOS, and RHEL.

* any time python-ldap would need a change in slapdtest helper, we would need to synchronize changes with you, wait until you have published a new version, upgrade dependencies, and finally do the change.
* we run the test suite in several stages of the Fedora, CentOS, and RHEL build and release process. I would have to get your package into Fedora (minor annoyance but doable, takes a couple of days and an hour or two of work time), but also into CentOS streams and RHEL buildroot (a **lot** of work, takes multiple weeks). Other Linux distros would either complicate their setup or they may choose to not run any tests.

On the other hand I don't see any apparent benefit from using an external package here.

azmeuk:
Thank you for your answer. I understand your point, I just wanted to make suggest the change as it has been discussed in another ticket.

> any time python-ldap would need a change in slapdtest helper, we would need to synchronize changes with you, wait until you have published a new version, upgrade dependencies, and finally do the change.

Not especially with me. python-slap is part of the python-ldap group, you have rw rights on the repository :)

PR #416, opened 2021-06-28
Fix SASL get/set options on big endian platforms

tiran:
The options OPT_X_SASL_SSF_MIN, OPT_X_SASL_SSF_MAX, and OPT_X_SASL_SSF
take *ber_len_t as input and output arguments. ber_len_t is defined as
unsigned long:

```
    /* LBER lengths (32 bits or larger) */
    #define LBER_LEN_T long

    typedef unsigned LBER_LEN_T ber_len_t;
```

Wrong type handling is causing issues on big endian platforms.

### DIFF
--- a/LICENCE.MIT
+++ b/LICENCE.MIT
@@ -6,6 +6,7 @@ to all contributions by the following authors:
 * Aymeric Augustin
 * Bernhard M. Wiedemann
 * Bradley Baetz
+* Christian Heimes
 * Éloi Rivard
 * Eyal Cherevatzki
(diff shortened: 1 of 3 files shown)


encukou:
LGTM.
Could you rebase to run with the latest CI config?
Also, add yourself to https://github.com/python-ldap/python-ldap/blob/master/LICENCE.MIT if you wish – I figure it's better if you do that explicitly.

droideck (approved):
LGTM too!

Issue #418, opened 2021-06-29
Restore all CI checks


encukou:
The tests ran at travis-ci.org were more comprehensive than what is run now with GitHub Actions.
We are missing:
* CFLAGS:  `-Wno-int-in-bool-context -Werror -Werror=declaration-after-statement`, `-std=c90` and `-std=c99` for Python 3.8+
* The `nosasltls` environment
* The `py3-trace` environment
* Coverage builds
* The docs build
* macOS build

We are also missing a Windows build.

spaceone:
Does anybody have a clue how to get `libldap` (or a fake version of it) correctly installed on MacOS in a Github Actions pipeline?

PR #419, opened 2021-06-29
Clean up tox environments & run tox on Fedora in CI 

encukou:
- Add CFLAGS used on Travis CI (`-Werror`, `std=c99` and similar) to tox environments
- Add `std=c90` environments for Python versions that compile with c90
- Add an environment for pypy3 and for Python 3.10 (a bit early, but unlikely to break now)
- Some general clean-up (remove python2 remnants, fix comments)
- Add tox-on-Fedora runs to the CI

Partially addresses: https://github.com/python-ldap/python-ldap/issues/418

~~The actions might not run on the PR that adds them, but they passed here: https://github.com/encukou/python-ldap/actions/runs/983127874~~

### DIFF
--- a/.github/workflows/tox-fedora.yml
+++ b/.github/workflows/tox-fedora.yml
@@ -1,0 +1,34 @@
+on: [push, pull_request]
+
+name: Tox on Fedora
+
+jobs:
+  tox_test:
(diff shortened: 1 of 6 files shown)


tiran (approved):
Nice!

PR #421, opened 2021-08-03
Add OPT_X_TLS_REQUIRE_SAN

tiran:
Add bindings for OPT_X_TLS_REQUIRE_SAN option. The flag was introduced
in OpenLDAP 2.4.52 to configure subject alternative name verification.


droideck (approved):
LGTM!

PR #401, opened 2021-02-25
support OPT_X_TLS_PEERCERT

graingert:

PR #427, opened 2021-09-15
Support OPT_X_TLS_PEERCERT

tiran:
Replaces PR https://github.com/python-ldap/python-ldap/pull/401

### DIFF
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -3,9 +3,11 @@ Automatic tests for python-ldap's module ldap.ldapobject
 
 See https://www.python-ldap.org/ for details.
 """
+import base64
 import errno
 import linecache
 import os
+import re
 import socket
 import unittest
 import pickle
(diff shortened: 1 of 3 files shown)


tiran:
Due to a bug in OpenLDAP, LDAP_OPT_X_TLS_PEERCERT leaks memory: https://bugs.openldap.org/show_bug.cgi?id=9696

Tests/t_ldapobject.py
```diff
@@ -418,6 +423,36 @@ def test_multiple_starttls(self):
...
+        peercert = l.get_option(ldap.OPT_X_TLS_PEERCERT)
+        self.assertTrue(peercert)
+        self.assertIsInstance(peercert, bytes)
+
+        if PEM_cert_to_DER_cert is not None:
+            with open(self.server.servercert) as f:
+                server_pem = f.read()
+            # remove text
+            begin = server_pem.find("-----BEGIN CERTIFICATE-----")
+            server_pem = server_pem[begin:-1]
```
encukou, line 451: nitpick: Is removing the final character important? In our case itdoesn't really matter, but if the file doesn't end with a newline, the exception you get here is a bit obscure.
```suggestion
            server_pem = server_pem[begin:]
```

Tests/t_ldapobject.py
```diff
@@ -421,6 +428,33 @@ def test_multiple_starttls(self):
             l.simple_bind_s(self.server.root_dn, self.server.root_pw)
             self.assertEqual(l.whoami_s(), 'dn:' + self.server.root_dn)
 
+    @requires_tls()
+    @unittest.skipUnless(
+        hasattr(ldap, "OPT_X_TLS_PEERCERT"),
```
mistotebe, line 433: Sounds like you still need to register this in `Lib/ldap/constants.py` to get the tests enabled?

mistotebe:
Already done as part of #458

Issue #423, opened 2021-08-30
set_option() does not handle OPT_X_SASL_SSF_EXTERNAL


tiran:
python-ldap exports ``OPT_X_SASL_SSF_EXTERNAL`` but there is no code that handles ``OPT_X_SASL_SSF_EXTERNAL``.

I'll add code and tests after #416 has been merged.

PR #428, opened 2021-09-15
Implement OPT_X_SASL_SSF_EXTERNAL setter

tiran:
The option flag ``OPT_X_SASL_SSF_EXTERNAL`` never worked because the
set_option code didn't handle the flag correctly.

Fixes: https://github.com/python-ldap/python-ldap/issues/423

### DIFF
--- a/Modules/options.c
+++ b/Modules/options.c
@@ -115,6 +115,7 @@ LDAP_set_option(LDAPObject *self, int option, PyObject *value)
 #ifdef HAVE_SASL
     case LDAP_OPT_X_SASL_SSF_MIN:
     case LDAP_OPT_X_SASL_SSF_MAX:
+    case LDAP_OPT_X_SASL_SSF_EXTERNAL:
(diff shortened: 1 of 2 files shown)


Modules/options.c
```diff
@@ -261,6 +262,12 @@ LDAP_get_option(LDAPObject *self, int option)
     Py_ssize_t i, num_extensions;
 
     switch (option) {
+#ifdef HAVE_SASL
+    case LDAP_OPT_X_SASL_SSF_EXTERNAL:
+        /* Read-only options */
```
encukou, line 267:
```suggestion
        /* Write-only option */
```

Modules/options.c
```diff
@@ -261,6 +262,12 @@ LDAP_get_option(LDAPObject *self, int option)
...
+        PyErr_SetString(PyExc_ValueError, "write-only option");
```
encukou, line 268: Could you also test that this error is raised?

tiran: Done

PR #430, opened 2021-09-17
Add another batch of names to LICENCE.MIT

encukou:
3 people since: https://github.com/python-ldap/python-ldap/pull/417#issuecomment-877068554


tiran (approved):

quanah:
And the term license isn't even used consistently in the documentation:

```
All contributions committed since July 1st, 2021, as well as some past contributions, are licensed under the MIT license. The MIT licence and more details are listed in the file LICENCE.MIT.
```

I.e., MIT license and MIT licence are both used in subsequent sentences.

PR #439, opened 2021-11-02
Fix wrong asserts in test_test_flags

jirutka:
Introduced in c5ad8025632fb5e5e84cced5a0f8c7ddda1e8dae

These conditions are redundant now, so I assume it’s a mistake.


droideck (approved):
Looks valid! Thanks!
You have my ack.

mistotebe (approved):
LGTM

PR #441, opened 2021-11-07
Update BooleanControl encoding for Python 3

antbob:
... to actually produce bytes as required. Not much else to say here except I tested this with LDAP Subentries request control (RFC 3672).

### DIFF
--- a/Lib/ldap/controls/simple.py
+++ b/Lib/ldap/controls/simple.py
@@ -57,8 +57,8 @@ class BooleanControl(LDAPControl):
   booleanValue
     Boolean (True/False or 1/0) which is the boolean controlValue.
   """
-  boolean2ber = { 1:'\x01\x01\xFF', 0:'\x01\x01\x00' }
-  ber2boolean = { '\x01\x01\xFF':1, '\x01\x01\x00':0 }
+  boolean2ber = { 1:b'\x01\x01\xFF', 0:b'\x01\x01\x00' }
+  ber2boolean = { b'\x01\x01\xFF':1, b'\x01\x01\x00':0 }
 
   def __init__(self,controlType=None,criticality=False,booleanValue=False):
     self.controlType = controlType


droideck (approved):
LGTM!

encukou:
Is there a good way to test this?

antbob:
@encukou For an API unit test you should do encode feeds decode dumb test as minimum. Also I see that you spawn OpenLDAP instance so if you want a real test thats the best way to go ie test against a different codebase not subject to any changes in your codebase. You can pick any simple control to do that. The one I have used you can see here

[ ](https://github.com/389ds/389-ds-base/blob/8a601ce8a9868ee18558eea1a359b1b42a72e74e/dirsrvtests/tests/suites/subentries/subentries_test.py)

Note that it hasn't been committed to 389ds so it wont work with it but OpenLDAP would work.

mistotebe:
Why is this file even encoding things manually when the rest of the codebase uses pyasn1? I thought that a `TRUE` can be encoded as any "integer != 0" in BER with `0xFF` being preferred, so we'd need to accept those, but let me know if I'm wrong.

antbob:
@mistotebe Yeah there are like 255 ways to ber encode boolean:true but generally its 0xff. I have no idea why you hardcode some encoding/decoding if you can use a library for that. Aside from some corner cases generally the only benefit would be perf related but in this case it wont make any practical diff, i dont think.

mistotebe:
> @mistotebe Yeah there are like 255 ways to ber encode boolean:true but generally its 0xff. I have no idea why you hardcode some encoding/decoding if you can use a library for that. Aside from some corner cases generally the only benefit would be perf related but in this case it wont make any practical diff, i dont think.

Why I'm thinking that rather than updating the value types, it should be fixed to do the right thing?

antbob:
@mistotebe You are thinking the right thing however I simply submitted the patch to get it working again but what you guys do with it is up to you.

droideck:
I think we should merge it as it is, and we can open a separate issue for improving this part of code (which will be a low hanging fruit for anyone).
@mistotebe what do you think?

P.S. the fix is needed for https://github.com/389ds/389-ds-base/pull/4985 to work (currently, it implements a workaround because of the bug)

mistotebe:
> I think we should merge it as it is, and we can open a separate issue for improving this part of code (which will be a low hanging fruit for anyone). @mistotebe what do you think?
> 
> P.S. the fix is needed for [389ds/389-ds-base#4985](https://github.com/389ds/389-ds-base/pull/4985) to work (currently, it implements a workaround because of the bug)

Can you test with https://github.com/mistotebe/python-ldap/commit/2db16e284ced98a2ef43362a0a949c3a10d5d35b then?

droideck:
> > I think we should merge it as it is, and we can open a separate issue for improving this part of code (which will be a low hanging fruit for anyone). @mistotebe what do you think?
> > P.S. the fix is needed for [389ds/389-ds-base#4985](https://github.com/389ds/389-ds-base/pull/4985) to work (currently, it implements a workaround because of the bug)
> 
> Can you test with [mistotebe@2db16e2](https://github.com/mistotebe/python-ldap/commit/2db16e284ced98a2ef43362a0a949c3a10d5d35b) then?

It works like a charm!
Then, I think we can abandon this PR. Feel free to open PR with your code, and I'll swiftly ack it.

mistotebe:
Superseded by #452

PR #451, opened 2022-01-16
Update installing.rst to reflect new openLDAP 2.5.0 is not yet supported

cfinnberg:
https://github.com/python-ldap/python-ldap/issues/432


mistotebe:
This is no longer relevant since 3.4.1 will be ok with OpenLDAP 2.5

PR #453, opened 2022-01-24
ldap.schema.models.Entry - Fix init()

Alphix:
I'm guessing this is partly due to commit
2be2c6bb8f44792d506cb1759fe06ff4cda3c5cf

And the update() method can't ever have worked?


mistotebe (approved):
LGTM

PR #456, opened 2022-01-26
Document TLS options do not apply immediately

mistotebe:
Until #55 has a better resolution, document the requirement to call `set_option( OPT_X_TLS_NEWCTX, 0 )` manually.


mistotebe:
Actually, scratch that, this is documented below already.

PR #458, opened 2022-01-27
Some detection whether we need to use -lldap_r to get threadsafe libldap

### LINKED ISSUE #432: Support for OpenLDAP 2.5+

mistotebe:
Closes #432

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ license_file = LICENCE
 
 # These defines needs OpenLDAP built with
 # ./configure --with-cyrus-sasl --with-tls
-defines = HAVE_SASL HAVE_TLS HAVE_LIBLDAP_R
+defines = HAVE_SASL HAVE_TLS
 
 extra_compile_args =
 extra_objects =
--- a/Doc/reference/ldap.rst
+++ b/Doc/reference/ldap.rst
@@ -349,7 +349,7 @@ TLS options
 .. py:data:: OPT_X_TLS_REQUIRE_SAN
 
    get/set how OpenLDAP validates subject alternative name extension,
-   available in OpenSSL 2.4.52 and newer.
+   available in OpenLDAP 2.4.52 and newer.
 
    :py:const:`OPT_X_TLS_NEVER`
       Don't check SAN
@@ -422,7 +422,7 @@ TLS options
 .. py:data:: OPT_X_TLS_PROTOCOL_MAX
 
    get/set maximum protocol version (wire protocol version as int),
-   available in OpenSSL 2.5 and newer.
+   available in OpenLDAP 2.5 and newer.
 
    .. versionadded:: 3.4.1
 


tiran (changes requested):
I recommend against using ctypes. The ``find_library`` function does not take distutils / setuptools options and flags into account. OpenLDAP 2.6 finally ships pkg-config files, so you could use pkg-config on Unix-like systems and figure out a fallback for OpenLDAP 2.4 and for Windows.

Or you could hook into build_ext command and use the compilers ``find_library_file`` helper to check for presence of ldap or ldap_r shared libs.

```
from setuptools.command.build_ext import build_ext

class ldap_build_ext(build_ext):
  def build_extension(self, ext):
    library_dirs = self.compiler.library_dirs + ext.library_dirs
    if self.compiler.find_library_file(library_dirs, "ldap_r"):
      ...
    super().build_extension(ext)
```

```
setup(
    ...,
    cmdclass = {'build_ext': ldap_build_ext},
    ...
)
```

I like the ``OPENLDAP_THREAD_SAFE`` check. Should we check the flags in the module initializer of the C extension instead?

mistotebe:
> I recommend against using ctypes. The `find_library` function does not take distutils / setuptools options and flags into account. OpenLDAP 2.6 finally ships pkg-config files, so you could use pkg-config on Unix-like systems and figure out a fallback for OpenLDAP 2.4 and for Windows.

Thanks for pointing me to setuptools API, didn't know these things were available, still nothing there I can find for examining pkg-config files nor tools to examine `ldap.h` to get the library version/feature flags so that doesn't get us all the way.

I could tweak the patch as you suggest below (using `cmdclass`) to find the right library path, but will have to `dlopen` it with ctypes anyway. Feel free to tweak the PR or suggest an alternative approach.

> I like the `OPENLDAP_THREAD_SAFE` check. Should we check the flags in the module initializer of the C extension instead?

Alternatively, we could just say we assume libldap is thread safe (switching to setting `LIBLDAP_R` at load time) and require that people building on distros where libldap and libldap_r do not point to the same library (if there are any left) choose to link to libldap_r by hand.

tiran:
setuptools has no code to run pkg-config. You need to role your own code, e.g. run ``pkg-config --libs``, shlex the output, and extend extra_linker_args. Do the same for cflags and extra compiler args.

You could use ``self.compiler.preprocess`` to run the pre-processor with a feature check file like this and check the exit code:

```
#include <ldap.h>

#ifndef LDAP_API_FEATURE_X_OPENLDAP_THREAD_SAFE
#errror "no thread safe"
#endif
```

Or keep it really simple:
- check for ``ldap_r``, if present use ``ldap_r``
- otherwise use ``ldap``
- add a check for thread safe libldap to ``_ldap`` extension module and fail to compile / load when a non-reentrant LDAP lib is present.

mistotebe:
Just been testing `find_library_file()` and if passed empty `library_dirs` (the default in our case), it will always return None anyway, so that's no use whatsoever for us.

tiran:
> Just been testing `find_library_file()` and if passed empty `library_dirs` (the default in our case), it will always return None anyway, so that's no use whatsoever for us.

As shown in my initial example, you have to pass a combination of compiler and extension library dirs to the function.

```
library_dirs = self.compiler.library_dirs + ext.library_dirs
```

mistotebe:
> As shown in my initial example, you have to pass a combination of compiler and extension library dirs to the function.

Which is what I tested with. Anyway, I did away with all that and just moved to the runtime check.

Modules/common.h
```diff
@@ -16,7 +16,7 @@
 #include <ldap.h>
 #include <ldap_features.h>
 
-#if LDAP_API_VERSION < 2040
+#if LDAP_VENDOR_VERSION < 20400
```
tiran, line 19: Good catch!

Modules/constants_generated.h
```diff
@@ -192,6 +194,7 @@ add_int(OPT_URI);
 add_int(OPT_DEFBASE);
 #endif
 
+
```
tiran, line 197: Where are the additional newlines coming from? The file is auto-generated by ``Lib/ldap/constants.py`` and Makefile.

setup.cfg
```diff
@@ -21,9 +21,11 @@ defines = HAVE_SASL HAVE_TLS HAVE_LIBLDAP_R
 extra_compile_args =
 extra_objects =
 
+# Uncomment this if your libldap is not thread-safe and you need libldap_r
+# instead
```
tiran, line 25: setup.cfg changes seem unrelated to this PR

Modules/constants_generated.h
mistotebe: I just ran the python script to regenerate, maybe that script changed in the meantime.

setup.cfg
mistotebe: If we keep the line there, people's builds with 2.5+ will keep failing, right?

mistotebe:
@tiran any feedback on the comments?

droideck (commented):
Besides the minor issue - looks good!

droideck (approved):
LGTM!
Thanks!

PR #459, opened 2022-01-31
Add 'referrals' to the LDAPError exception as a list containing all referrals returned from server

nickfolino:
Added a refList that contains all referrals returned from the server.
Still maintains the first referral in 'info'.

### DIFF
--- a/Modules/constants.c
+++ b/Modules/constants.c
@@ -165,10 +165,32 @@ LDAPraise_for_message(LDAP *l, LDAPMessage *m)
         if (errnum == LDAP_REFERRAL && refs != NULL && refs[0] != NULL) {
             /* Keep old behaviour, overshadow error message */
             char err[1024];
+            PyObject *referralList;
+            int i;
 
+            for ( i=0; refs[i]; i++ ) /* count */;
+            referralList = PyList_New(i);
+            if (referralList) {
+                for ( ; i--; ) {
+                    PyObject *referralURL = Py_BuildValue("s", refs[i]);
+                    if (referralURL == NULL ||
+                            PyList_SetItem(referralList, i, referralURL)) {
+                        Py_CLEAR(referralList);
+                        break;
+                    }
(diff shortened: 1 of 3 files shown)


Modules/constants.c
```diff
@@ -162,9 +162,18 @@ LDAPraise_for_message(LDAP *l, LDAPMessage *m)
         }
 
         if (errnum == LDAP_REFERRAL && refs != NULL && refs[0] != NULL) {
+            /* Return all referrals as a list in "reflist" */
+            PyObject *referralList = PyList_New(0);
```
tiran, line 166: Missing error check.

Modules/constants.c
```diff
@@ -162,9 +162,18 @@ LDAPraise_for_message(LDAP *l, LDAPMessage *m)
...
+            int i = 0;
+            while (refs[i] != NULL) {
+                PyObject *referralURL = Py_BuildValue("s", refs[i++]);
```
tiran, line 169: Use ``PyUnicode_FromString()`` and check for error (NULL).

Modules/constants.c
```diff
@@ -162,9 +162,18 @@ LDAPraise_for_message(LDAP *l, LDAPMessage *m)
...
+                PyList_Append(referralList, referralURL);
```
tiran, line 170: Missing error check.

Modules/constants.c
```diff
@@ -162,9 +162,18 @@ LDAPraise_for_message(LDAP *l, LDAPMessage *m)
...
+                Py_XDECREF(referralURL);
+            }
+            PyDict_SetItemString(info, "reflist", referralList);
```
tiran, line 173: Missing error check

tiran (dismissed):
Every Python API call requires an explicit error check.

Please also include documentation update and test cases.

droideck (approved):
LGTM!

Issue #448, opened 2021-12-10
Can't use ReconnectLDAPObject to bind because of TypeError


jnewbigin:
Issue description:
Using `ReconnectLDAPObject` results in an error:
`TypeError: _store_last_bind() got multiple values for argument 'method'`

Steps to reproduce: 

```
connection = ldap.ldapobject.ReconnectLDAPObject(uri=uri, bytes_mode=False)
connection.bind_s(bind_dn, password, method=ldap.AUTH_SIMPLE)
```

`bind_s` invokes `self._store_last_bind`

```
self._store_last_bind(method=SimpleLDAPObject.bind_s,*('username', 'password'),**{'method': 128})
```
Which has 2 parameters called method and hence the exception.

In my case, I can omit the `method=ldap.AUTH_SIMPLE` but if I needed a different auth method then it would not work.

Operating system: 
Linux

Python version: 
3.8.11

python-ldap version: 
python-ldap==3.3.1

mistotebe:
I guess it's a good thing LDAP never really used a different AuthenticationChoice than `simple` and `sasl`, so noone has a reason to use the method parameter.

In any case, the first argument to `_store_last_bind` should be renamed to something else since that's an internal API and something we can fix without breaking users.

jnewbigin:
I think we were just being extra careful with our original code and specifying everything even though the defaults were OK.

mistotebe:
> I think we were just being extra careful with our original code and specifying everything even though the defaults were OK.

Hi @jnewbigin, could you confirm #471 works as expected?
Thanks

jnewbigin:
:+1: works great thanks.

Issue #67, opened 2017-11-30
Check LDAP_OPT constants


tiran:
Double check that we provide constants for all supported ``OPT`` constants and have documentation for all of them. We might want to decide to NOT expose some constants, e.g. callbacks or getters that return raw C structures like ``OPT_X_TLS_CTX``.

See http://man7.org/linux/man-pages/man3/ldap_get_option.3.html for an up to date list 

## missing / unsupported

* [ ] OPT_X_TLS_ECNAME
* [x] OPT_X_TLS_PEERCERT
* [ ] OPT_X_TLS_PROTOCOL_SSL3
* [ ] OPT_X_TLS_PROTOCOL_TLS1_0
* [ ] OPT_X_TLS_PROTOCOL_TLS1_1
* [ ] OPT_X_TLS_PROTOCOL_TLS1_2
* [ ] OPT_X_TLS_PROTOCOL_TLS1_3

## drop / don't support

* [ ] OPT_X_TLS_PROTOCOL_SSL2
* [ ] OPT_X_TLS_PROTOCOL_SSL3
* [ ] OPT_X_TLS_TRY makes no sense. It's only useful on the server side.
* [ ] OPT_X_TLS_CTX
* [ ] OPT_X_TLS is deprecated and no longer documented in OpenLDAP: https://www.openldap.org/lists/openldap-software/200409/msg00617.html
* [x] OPT_X_*_CB (C callback)
* [x] OPT_X_*_ARG (argument for C callback)

PR #461, opened 2022-02-14
Process missing libldap options

mistotebe:


mistotebe:
I've integrated #427 with my comment resolved. Due to changes to the generated header, I've also included #458, don't know how to mark it as a dependency.

Doc/reference/ldap.rst
```diff
@@ -406,7 +406,12 @@ TLS options
 
 .. py:data:: OPT_X_TLS_PEERCERT
 
-   Get peer's certificate as binary ASN.1 data structure (not supported)
+   Get peer's certificate as binary ASN.1 data structure (DER)
+
+   .. versionadded:: 3.4.0
```
droideck, line 411: If I understand our release process correctly, the right release version will be 3.4.1 here.

3.4.0 is already release and it doesn't have OPT_X_TLS_PEERCERT option.

droideck (approved):
Besides the minor issue, looks good!

Doc/reference/ldap.rst
mistotebe: Hmm, should we use a placeholder in these and leave replacing that to the release process?

droideck: We can do that, but I think it makes more sense to plan the changes because if we use a placeholder, then when we have the next release, it will be dynamically filled with some amount of fixes/features, and it'll feel too unpredictable.

So I think it'll be more natural to triage the issues we work on, assign the milestone, and then when we have the next release, we do the tagging, etc.
Also, this way we'll be able to plan major changes ahead of time (like the removal of Python 2 was planned for 3.4.0).

mistotebe: I've updated this, also removed some deprecated options and constants according to #67. I guess if everything is OK, we'll get OpenLDAP 2.5 support and that might warrant pushing out 3.4.1, unless we want some other things in there too, creating a milestone for tracking.

droideck: I agree that OpenLDAP 2.5 support is already a solid reason to release. So as it's done, I think, we can review what has changed since 3.4.0 release and push out 3.4.1 as you suggested. :)

droideck (approved):
I've run a few tests and it looks great!
The code looks good too! Ack.

Doc/reference/ldap.rst
mistotebe: Thanks, I've merged this. Could you review what other issues should go into 3.4.1 and add them? Also review pending PRs, I think I'll add a fix for #448 into that shortly.

PR #463, opened 2022-03-09
SlapdObject updates

mdavidsaver:
A three part change to `SlapdObject`, which I've found helpful in testing a project using python-ldap.

1. Look for `slapadd` under `/usr/sbin`, where [Debian](https://packages.debian.org/file:slapadd) puts it.
2. Bind to `ldaps://` as well.
3. Include a mechanism for injecting `olcAccess` rules.

The second part turned out to be a more involved change due to some race conditions turning up in CI runs.  I think I've resolved these by expanding `_avail_tcp_ports()` to poll for multiple unused TCP ports, and adjusting `_start_slapd()` to wait for a successful `ldapwhoami()` against all listening sockets.


Lib/slapdtest/_slapdtest.py
```diff
@@ -259,7 +275,7 @@ def _find_commands(self):
         self.PATH_LDAPDELETE = self._find_command('ldapdelete')
         self.PATH_LDAPMODIFY = self._find_command('ldapmodify')
         self.PATH_LDAPWHOAMI = self._find_command('ldapwhoami')
-        self.PATH_SLAPADD = self._find_command('slapadd')
+        self.PATH_SLAPADD = self._find_command('slapadd', in_sbin=True)
```
mistotebe, line 278: Can you switch to doing the same as #282 did?

Lib/slapdtest/_slapdtest.py
```diff
@@ -347,6 +380,7 @@ def gen_config(self):
             'cafile': self.cafile,
             'servercert': self.servercert,
             'serverkey': self.serverkey,
+            'access': '', # '\n'.join(self.gen_access()),
```
mistotebe, line 383: Why is this commented out?

Lib/slapdtest/_slapdtest.py
```diff
@@ -425,13 +459,18 @@ def _start_slapd(self):
                 self._log.debug(
                     "slapd connection check to %s", self.default_ldap_uri
                 )
-                self.ldapwhoami()
+                # try all bound sockets to make sure the daemon is ready to go.
+                self.ldapwhoami(ldap_uri=self.ldap_uri)
+                self.ldapwhoami(ldap_uri=self.ldaps_uri)
+                if self.ldapi_uri:
+                    self.ldapwhoami(ldap_uri=self.ldapi_uri)
```
mistotebe, line 466: If one uri is contactable, the others will too, no need to check all.

Lib/slapdtest/_slapdtest.py
```diff
@@ -496,16 +535,36 @@ def _cli_auth_args(self):
...
             ldap_uri = self.default_ldap_uri
 
+        environ = os.environ.copy()
+
         if ldapcommand.split("/")[-1].startswith("ldap"):
-            args = [ldapcommand, '-H', ldap_uri] + self._cli_auth_args()
+            args = [ldapcommand] + self._cli_remote_args(ldap_uri)
+
+            # setting $LDAPNOINIT ignores all other environment and config files
+            environ.pop('LDAPNOINIT', None)
```
mistotebe, line 552: This is a -1 from me because you would need to enumerate *all* libldap options here so that people's ldap.conf doesn't leak in as you rightly noticed. You'll need to use `-o` command-line options to provide tls settings (not in 2.4 unfortunately).

Lib/slapdtest/_slapdtest.py
```diff
@@ -314,18 +330,35 @@ def _cleanup_rundir(self):
...
         """
-        sock = socket.socket()
-        try:
-            sock.bind((self.local_host, 0))
-            port = sock.getsockname()[1]
-        finally:
-            sock.close()
-        self._log.info('Found available port %d', port)
-        return port
+        return self._avail_tcp_ports(1)
```
mistotebe, line 352: Aren't you changing the return type from `int` to `list[int]` here?

mistotebe (changes requested):

Lib/slapdtest/_slapdtest.py
mdavidsaver: A mistake on my part.

Lib/slapdtest/_slapdtest.py
mdavidsaver: I was actually seeing sporadic failures on some GHA runs which appear to have stopped with this addition.  The specific failure from `pypy3` with a `Bad file descriptor` during `tearDown()` after `Test03_SimpleLDAPObjectWithFileno.test_attrlist_accepts_unicode`.  You might be able to [see this log](https://github.com/mdavidsaver/python-ldap/runs/5470738478?check_suite_focus=true#step:7:77).

I didn't dig into the specifics of this test failure as I'm already digressing from my project, and I had an intuition that incorrect startup was involved.  Maybe the test is swallowing a `connect()` failure?

Consider that the `slapd` process is not instantaneously/atomically calling `bind()` and `accept()` on all sockets.  So there will be a period of time, however short, when some sockets can be `connect()`ed, and others can't.  I know this seems unlikely, but I've found that CI has a way of triggering these sorts of races.

Lib/slapdtest/_slapdtest.py
mdavidsaver: I didn't notice `-o <opt>=<val>`.  I agree that this is preferable.  Any idea what the `<opt>` name is for `TLS_CACERT`?  I tried some variations on `-o tls-cacert=...` without success.

[`man ldapadd`](https://linux.die.net/man/1/ldapadd) gives an example of `-o nettimeout=...`, which I guess corresponds to `NETWORK_TIMEOUT` in [`ldap.conf`](https://linux.die.net/man/5/ldap.conf).  So there doesn't seem to be a programmatic mapping.

quanah: In the OpenLDAP test suite, test068 uses a number of the -o options.  If you examine it you will find the following:

```
$LDAPSASLWHOAMI -o tls_cacert=$TESTDIR/tls/ca/certs/testsuiteCA.crt -o tls_reqcert=hard \
    -o tls_cert=$TESTDIR/tls/certs/bjensen@mailgw.example.com.crt -o tls_key=$TESTDIR/tls/private/bjensen@mailgw.example.com.key -ZZ -Y EXTERNAL -H $URIP1 \
    > $TESTOUT 2>&1
```

which indicates that an underscore is used as the separator rather than a hyphen.  Hope that helps!

quanah: Also, the reason nettimeout is different is that it was implemented singularly before the global handling of options was added for 2.5. :/  Be nice to make it consistent, but that would break backwards compatibility.

mdavidsaver: In case others find this.  As I read it, the authoritative list of option names in the openldap tree is [`libraries/libldap/init.c`](https://git.openldap.org/openldap/openldap/-/blob/07429efcff9dc50abb8ca265c8e8caa8cd25d01c/libraries/libldap/init.c#L80) with the ~legacy CLI-only synonyms in [`clients/tools/common.c`](https://git.openldap.org/openldap/openldap/-/blob/07429efcff9dc50abb8ca265c8e8caa8cd25d01c/clients/tools/common.c#L821).  Option names are matched with `strcasecmp()` (case insensitive) after [replacing](https://git.openldap.org/openldap/openldap/-/blob/07429efcff9dc50abb8ca265c8e8caa8cd25d01c/clients/tools/common.c#L815) '-' with '_'.  So it seems like `-o TLS_CACERT=...` and `-o tls-cacert=...` should be equivalent.

It isn't immediately clear to me why my initial guesses didn't seem to work.  I guess I had some other change which I didn't notice...  I'm making my first deep dive into the details of ldap, and am still wading through all of the gotchas.

mdavidsaver: Right, `-o tls_cacert=` doesn't work for me because I'm still using the Debian [packaged](https://tracker.debian.org/pkg/openldap) openldap 2.4.57.  The expanded `-o` is added in 2.5.x.  So I think I'm going to scale back my ambitions.

mdavidsaver:
I've dropped the changes involving `ldaps://`, which moots most of the reviewer comments.  I did switch from `slapadd` to `slapd -Tadd` as requested.  I also added `bind_dn=` and `bind_pw=` to the `ldap*` methods.  eg. `ldapwhoami(bind_dn="CN=special,...", bind_pw="...")`.

Also added is a warning if `slapadd()` is called while the daemon is running.  This is only a warning because one test (`test_slapadd` in `Tests/t_ldapobject.py`) does this.  Although just before calling `restart()`, so it probably has no ill effects.

Lib/slapdtest/_slapdtest.py
```diff
@@ -483,34 +482,57 @@ def _stopped(self):
...
+        args = [ldapcommand, '-H', ldap_uri or self.default_ldap_uri]
+
+        if ldap_uri.startswith('ldapi://'):
+            args += ['-Y', 'EXTERNAL']
+
+            if not self._log.isEnabledFor(logging.DEBUG):
+                args.append('-Q')
+
+            if bind_dn or bind_pw:
+                raise RuntimeError('-Y EXTERNAL ignores -D and -w')
```
mistotebe, line 520: How about treating those being passed in as a signal to use a simple bind? Also the `self.cli_sasl_external` check is gone?

mdavidsaver: > How about treating those being passed in as a signal to use a simple bind?

Maybe showing my ignorance here.  Is a simple bind meaningful with `ldapi://`?

So the condition for choosing `-Y EXTERNAL` would be `ldap_uri.startswith('ldapi://') and bind_dn is None`?

> Also the `self.cli_sasl_external` check is gone?

Yes.  I don't think this was the correct condition.  Knowing that `ldapi://` is supported doesn't imply that it will be used for a particular operation.

mistotebe: The SASL `EXTERNAL` mechanism is independent of the `ldapi://` transport. `cli_sasl_external` tells you whether that mechanism is available/should be used, `ldapi://` tells you how to contact the server (over a UNIX socket vs. as opposed to a TCP socket), the LDAP protocol still behaves the same over both.

Hope that clears things up a bit, let me know if not.

mistotebe: For example if we set up a TLS layer and used a client certificate while doing so, some servers could also accept the `EXTERNAL` mechanism, using the client certificate to generate the identity the session should bind as (OpenLDAP does this).

BTW not asking that you take it into account in this pull request, just giving some background.

PR #464, opened 2022-03-24
Draft: new response handling API

mistotebe:
- [x] tests
- [x] intermediate/exop decoding

### DIFF
--- a/Lib/ldap/extop/__init__.py
+++ b/Lib/ldap/extop/__init__.py
@@ -10,6 +10,7 @@ response.
 """
 
 from ldap import __version__
+from ldap import KNOWN_EXTENDED_RESPONSES, KNOWN_INTERMEDIATE_RESPONSES
 
 
 class ExtendedRequest:
@@ -42,18 +43,45 @@ class ExtendedResponse:
   """
   Generic base class for a LDAPv3 extended operation response
 
-  requestName
-      OID as string of the LDAPv3 extended operation response
+  responseName
+      OID as string of the LDAPv3 extended operation response or None
   encodedResponseValue
       BER-encoded ASN.1 value of the LDAPv3 extended operation response
   """
+
+  def __init_subclass__(cls):
+    if not getattr(cls, 'responseName', None):
+      return
+
+    KNOWN_EXTENDED_RESPONSES.setdefault(cls.responseName, cls)
 
   def __init__(self,responseName,encodedResponseValue):
     self.responseName = responseName
     self.responseValue = self.decodeResponseValue(encodedResponseValue)
 
-  def __repr__(self):
-    return f'{self.__class__.__name__}({self.responseName},{self.responseValue})'
+  def decodeResponseValue(self,value):
+    """
+    decodes the BER-encoded ASN.1 extended operation response value and
+    sets the appropriate class attributes
+    """
+    return value
+
+
+class IntermediateResponse:
+  """
+  Generic base class for a LDAPv3 intermediate response message
+
+  responseName
+      OID as string of the LDAPv3 intermediate response message or None
+  encodedResponseValue
+      BER-encoded ASN.1 value of the LDAPv3 intermediate response message
+  """
+
+  def __init_subclass__(cls):
+    if not getattr(cls, 'responseName', None):
+      return
+
+    KNOWN_INTERMEDIATE_RESPONSES.setdefault(cls.responseName, cls)
 
   def decodeResponseValue(self,value):
     """
--- a/Lib/ldap/extop/__init__.py
+++ b/Lib/ldap/extop/__init__.py
@@ -12,6 +12,13 @@ response.
 from ldap import __version__
 from ldap import KNOWN_EXTENDED_RESPONSES, KNOWN_INTERMEDIATE_RESPONSES
 
+import ldap
+import ldap.response
+
+from typing import Optional
+
+_NOTSET = object()
+
 
 class ExtendedRequest:
   """
@@ -39,7 +46,7 @@ class ExtendedRequest:
     return self.requestValue
 
 
-class ExtendedResponse:
+class ExtendedResponse(ldap.response.ExtendedResult):
   """
   Generic base class for a LDAPv3 extended operation response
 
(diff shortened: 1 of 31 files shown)


Modules/LDAPObject.c
```diff
@@ -16,6 +16,33 @@
...
+    {
+        .name = NULL,
+    }
+};
+
+PyStructSequence_Desc message_tuple_desc = {
+    .name = "_RawLDAPMessage",
+    .doc = "LDAP Message returned from native code",
+    .fields = message_fields,
+    .n_in_sequence = sizeof(message_fields)/sizeof(*message_fields) - 1,
```
encukou, line 41: Set this to 2 to preserve backwards compatibility in `result`. (And ideally, at the same time, deprecate using `result()` as a tuple, along with <code>result<var>N</var></code>, in favor of the named fields).
This would allow old code to continue using `type, data = result()`.

Setting this to a value that will grow as more fields are added defeats the purpose of `PyStructSequence` – you might as well use a regular object with getset descriptors.

mistotebe: This doesn't feel remotely compatible with `LDAPObject.result()` anyway, that's one of the reasons the new API is split into a Connection object (which gets a better request API too in a future PR). Not sure whether to make the result* APIs a (heavyweight) wrapper for this or keep it for now.

Keeping `n_in_sequence` automatic is an oversight from initial development, thanks for pointing that out, I'll fix it in the next version. What is your take on the Response subclassing logic? I will probably have to replicate it for ExtendedResponse and intermediates, can you see any conceptual issues with how it is written/is it a total hack?

encukou: Sadly I won't have time to review the whole PR :(
Could you point to the part I should look at?

mistotebe: Turns out I forgot to add the Python files into the commit so my last comment probably didn't make much sense. I meant the Python bits, [this](https://github.com/python-ldap/python-ldap/pull/464/files#diff-204cac07a764dc89fd5053d3dce3b2a1f33d6b85387b4ef9a265ec39c1b06233R52-R70) in particular.

encukou: At first glance, that looks more complicated than it needs to be. i don't see much advantage over setting the missing attributes to None (or keeping them missing). Is there an advantage?

mistotebe: The subclassing allows one to expose response-specific methods (Compare result, allow parsing the extended result) and if the user is on 3.10, allow pattern matching, on others, make the code more readable, especially if controls were subjected to the same treatment.

droideck:
@mistotebe I'm thinking about python-ldap 4.0 release.
Do you think this PR is ready for review? We can include it earlier (as a tech preview or something like that). And while it's being improved and handled as an experimental feature, we can stabilize it for python-ldap 5.0 in the future.
Does it sound good to you?

mistotebe:
Any feedback is welcome, yes. I've been using this branch elsewhere with some success and the tools added here certainly make porting some of my tools (e.g. syncmonitor) from python-ldap0 an to some extent pldap easier. Don't think I'll be abandoning pldap any time soon (there is no replacement to its stub server implementation for protocol level testing) but definitely closer to dropping python-ldap0 after we get something like this in.

There's still a little bit of a kludge in how e.g. searches are represented, so the API might still evolve, and it's only a stepping stone towards:
- a more ergonomic `async`-friendly API
- a proper overhaul of the request side

But it's definitely useful already so happy to start cleaning up or breaking it up starting with bits you consider less contentious and go from there.

mistotebe:
> > a more ergonomic async-friendly API
> > a proper overhaul of the request side
> 
> Maybe my project is suitable for this. I am supporting a nice user-friendly API with sync and async, hiding the `_s` and `_ext` details but providing all functionality LDAP clients supports (on top of python-ldap) and especially focusing on correctness for example regarding DN and filter escape handling: See https://docs.freeiam.org/ and https://github.com/Free-IAM/freeiam.

You should definitely be able to make use of the new interfaces and response classes added here to make your own life easier.

In terms of request API (and async), yes yours looks somewhat similar to what I wanted, but there is one important difference: **any** request (not just search) can return any amount of "intermediate" responses (search entry, referral entry, intermediate) and then a final response or exception so a request object should (always?) be iterable. A caller might be able to say they (don't) expect those, they still have a choice to let the whole request finish or they can receive messages one by one and pattern match on the type (SearchResultEntry, SyncInfoNewCookie, generic IntermediateResponse, ...) etc.

There is a partial draft of [how the async side could look](https://git.openldap.org/openldap/syncmonitor/-/blob/master/syncmonitor/ldap_wrapper.py#L38), what you get is a Future and optionally an iterable for the intermediate responses.

We can definitely see how to converge those two APIs because in some ways your take is close to what I had in mind and a lot of things I can see there should simplify to "syntactic sugar"/shorthand over the new API.

spaceone:
Thanks for the explanation.
I already use provide iterated search results via `all=1` , but that seems to be a lot slower, so i renamed it to `search_iter()`.
For the referrals, I thought that wouldn't be a intermediate result but a final result. My plan was to add a connection pool, which also handles referrals.
So I will change my API that the result can be a future and a nice way to decide whether you want a iterable or not.
I also need to find out how I can trigger these intermediate responses with openldap, so I still can get realistic 100% test coverage.

mistotebe:
> Thanks for the explanation. I already use provide iterated search results via `all=1` , but that seems to be a lot slower, so i renamed it to `search_iter()`.

Slower? Are we both talking about ldap_result() with `all=MSG_ONE` (and if async, having a select() there)?

> For the referrals, I thought that wouldn't be a intermediate result but a final result. My plan was to add a connection pool, which also handles referrals.

I mean SearchReference, not Result.result == LDAP_REFERRAL. Opting to handle referrals is an option, sure, but the client should have an option to return them (and sometimes that makes sense?).

>  So I will change my API that the result can be a future and a nice way to decide whether you want a iterable or not. I also need to find out how I can trigger these intermediate responses with openldap, so I still can get realistic 100% test coverage.

Often syncrepl control (mode=refreshAndPersist) is the easiest one, or LDAP TXN (RFC 5805). Both exist in OpenLDAP, syncrepl needs the syncprov overlay configured, TXNs need a compatible backend (e.g. back-mdb).

Modules/LDAPObject.c
```diff
@@ -1035,6 +1062,437 @@ l_ldap_rename(LDAPObject *self, PyObject *args)
...
+                    if ( !pytmp || PyList_SetItem( value_list, index, pytmp ) ) {
+                        ldap_memfree( values );
+                        Py_DECREF(value_list);
+                        goto error;
+                    }
+                }
+                pytmp = NULL;
+
+                /* FIXME: deal with attrs-only */
+                assert(0);
```
droideck, line 1214: This, probably, should be completed before the merge

Lib/ldap/extop/__init__.py
```diff
@@ -38,19 +46,159 @@ def encodedRequestValue(self):
...
+    Supports both old and new API:
+    __init__(self, responseName: Optional[str],
+             encodedResponseValue: Optional[bytes])
+    and
+    __init__(self, msgid: int, msgtype: int, controls: Controls = None, *,
+             result: int, matcheddn: str, message: str, referrals: List[str],
+             defaultClass: Optional[type[ExtendedResponse]] = None,
+             **kwargs)
+
+    The old API is deprecated and will be removed in 4.0.
```
droideck, line 164: If you agree with my proposal, this version comment can be changed.

Lib/ldap/extop/__init__.py
```diff
@@ -38,19 +46,159 @@ def encodedRequestValue(self):
...
+                        responseName=_NOTSET, encodedResponseValue=_NOTSET,
+                        **kwargs):
+    """
+    Implements both old and new API:
+    __init__(self, responseName, encodedResponseValue)
+    and
+    __init__/__new__(self, msgid, msgtype, controls=None, *,
+                     result, matcheddn, message, referrals,
+                     defaultClass=None, **kwargs)
+    """
```
droideck, line 82: It looks overly complex, IMO.
Can we split it into two?

Something like
```python
@classmethod
def from_old_api(cls, responseName, encodedResponseValue)
@classmethod  
def from_new_api(cls, msgid, msgtype, ...)
```

droideck (commented):
While I can't see any obvious leaks, the complex control flow around error handling needs careful testing under various failure scenarios. We might need to have an ASAN run here.

And, of course, all tests need to pass consistently before we merge anything.

mistotebe:
Thanks for looking at it. Broadly I agree with the plan except that delaying Connection to 5.0 makes all the 4.0 additions moot because it leaves no way of reaching all the new code/classes.

How about this?
- 4.0 adds `Connection` class with a note that the API is experimental and **will** evolve during 4.x in incompatible ways (e.g. request side of the API, async support)
- 5.0 when we consider the Connection API stable and we can mark `LDAPObject` as deprecated (might even reimplement it as a wrapper of `Connection` while preserving behaviour)
- 6.0 LDAPObject and old APIs in the C module that aren't needed by Connection can disappear

Does all the mucking with `__new__` to redirect to the appropriate subclass feel like a hack? Is there a better scheme to achieve something similar? I think I tried using factories to do this, not sure why I abandoned that approach or if I just never finished hashing it out.

droideck (commented):
I was thinking about those components as standalone tools as I thought you want smaller scale of that exposed in 4.0.
But yeah, without Connection there's no easy way to actually use any of the new stuff. Your timeline makes more sense.

So 4.0 with experimental/tech preview Connection, 5.0 stable + deprecate LDAPObject, 6.0 remove old stuff - sounds good to me.

About the `__new__` pattern - It's clever but honestly it does feel a bit too magical... The main issues I see:

- Readability - When someone calls `Response(msgid, msgtype, ...)` and gets back a `SearchEntry`, that's surprising behavior, so it violates the principle of least surprise. The metaclass magic isn't immediately obvious when debugging.
- The class-level registry - That `__subclasses` dict could be problematic if someone ever subclasses Response for their own framework. Plus the assert in `__init_subclass__` is a bit aggressive (I left a comment below).

Alternative approaches that might be cleaner:

- Explicit factory functions: Just have `response_from_message(msgid, msgtype, ...)` that does the dispatch. Less magical, easier to debug.
- A proper factory class: Something like `ResponseFactory.create(msgid, msgtype, ...)`. You could even make it pluggable for custom response types.
- Keep __new__ but make it more explicit: Maybe rename Response to ResponseBase or AbstractResponse so it's clear it's not meant to be instantiated directly?

That said, if the pattern works and you're comfortable with it, I'm not too much against that. The automatic subclass registration via `__init_subclass__` is actually pretty neat. Maybe just needs better documentation/comments explaining the magic?

mistotebe:
It won't let me "reply" to [your comment on `__convert_old_api`](https://github.com/python-ldap/python-ldap/pull/464#discussion_r2288820040), so replying separately:

> Maybe we can create a temporary transitional class? Like ExtendedResultOrSomething(ldap.response.ExtendedResult): with the new implementation? And in 5.0, we can swap it with the existedExtendedResponse

I'm happy with anything of the sort as long as there's a way to reuse most of the decoding and registration code in `ldap.extop`. If we can find a way that gets rid of this as well, I'll only be happier, not less so.

Lib/ldap/extop/__init__.py
mistotebe: (Ah found it, there's several copies of this comment in the interface, copying my reply here)

I'm happy with anything of the sort as long as there's a way to reuse most of the decoding and registration code in `ldap.extop`. If we can find a way that gets rid of this as well, I'll only be happier, not less so.

mistotebe: I've tried changing the routing towards a `from_message` class method. This has removed some of the magic from the constructors but Extended ops are still going to need the scary variadic signature unless I missed something? Should be kept isolated in the last commit so it can be dropped if we don't like it.

mistotebe:
At the moment, there are two types of test failures:
- in pypy we still somehow close file descriptors when we shouldn't, this must be something non-deterministic about how pypy orders destructors or maybe we don't refcount properly? I've only reproduced this once on my machine after running this repeatedly and don't know what/how to trace things
- Python 3.9 complains that `ldap.SUCCESS` has no attribute `errnum`, but others are ok and I don't see anything version specific in `add_err` or elsewhere

I don't have an answer to either but I also don't think they are related to this PR.

PR #493, opened 2022-08-26
Add ASN.1 replace to psearch control

slyngshede:
Allow Sphinx to pickup the ASN.1 substitution from pyasn1, so docs can build.

I'm not completely sure that this is how Sphinx is suppose to work though.


slyngshede:
My initial local test didn't catch that the ASN.1 substitute is used in more than one place.

droideck (approved):
LGTM!

PR #494, opened 2022-08-26
Fix broken reset and tearDown in pypy3

slyngshede:
There appear to be some minor differences in the CPython and Pypy3 implementations of del. 
Use delattr to ensure that hasattr will see the attributes as deleted.

This should fix to of the failing pypy3 test cases.

### DIFF
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -658,12 +658,12 @@ class Test03_SimpleLDAPObjectWithFileno(Test00_SimpleLDAPObject):
 
     def tearDown(self):
         self._sock.close()
-        del self._sock
+        delattr(self, '_sock')
         super().tearDown()
 
     def reset_connection(self):
         self._sock.close()
-        del self._sock
+        delattr(self, '_sock')
         super(Test03_SimpleLDAPObjectWithFileno, self).reset_connection()
 
 


droideck:
Could you please rebase the PR with the ASN.1 fixes from the master branch?

And thank you so much for the commits! We really appreciate it!

slyngshede:
While I'm not entirely sure how or the rebase worked, it appears to have passed all the tests now. Please let me know if you prefer the rebase be done differently. 

Happy to help :-)

droideck (approved):
Thanks!
Looks good!

mistotebe:
> There appear to be some minor differences in the CPython and Pypy3 implementations of del.
> Use delattr to ensure that hasattr will see the attributes as deleted.

Could you explain why you think this is the case? A quick test doesn't support your suggestion:
```
$ pypy3
Python 3.8.13 (7.3.9+dfsg-4, Aug 09 2022, 12:51:24)
[PyPy 7.3.9 with GCC 12.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>> class A:
....     def __init__(self):
....         self.a = 1
....         self._b = 2
....         
>>>> a = A()
>>>> a
<__main__.A object at 0x00007f3fc525b0c0>
>>>> [hasattr(a, name) for name in ['a', '_b', 'c']]
[True, True, False]
>>>> a.c = 3
>>>> [hasattr(a, name) for name in ['a', '_b', 'c']]
[True, True, True]
>>>> del a.c
>>>> [hasattr(a, name) for name in ['a', '_b', 'c']]
[True, True, False]
>>>> del a._b
>>>> [hasattr(a, name) for name in ['a', '_b', 'c']]
[True, False, False]
```

slyngshede:
I am by no means certain that this is correct, but the tests I did with the pypy3 tests cases appeared to fail on the hasattr check. With the attribute gone, the connection should reset. There are no other obvious code path it could take. 

If I recall, I can't find the failing test now, this is where the test failed:
```
def _open_ldap_conn(self, who=None, cred=None, **kwargs):
        if hasattr(self, '_sock'):
            raise RuntimeError("socket already connected")
        self._sock = socket.create_connection(
            (self.server.hostname, self.server.port)
        )
        return super()._open_ldap_conn(
            who=who, cred=cred, fileno=self._sock.fileno(), **kwargs
        )
```

New connection should have been created, there are no other code paths, except if the '_sock' attribute exits. So clearly the '_sock' attribute isn't deleted. Both del and delattr should do the same, in fact del should be a bit more efficient. 

The exact reason may be somewhere in the pypy3 code, my guess is that the GC hasn't run, so the object is technically still around. delattr is either slow enough or simply ensures that the objects is truly deleted.

I can try looking at the pypy3 source a bit later and see if the two functions are implemented differently.

PR #500, opened 2022-10-27
Add Python 3.10 and 3.11 to CI and tox

joshuadavidthomas:
Noticed Python 3.10 was missing from GitHub Actions and trove classifiers and 3.11 was missing from those two plus tox.

Stylistic nit: in GHA, quotes are needed around 3.10 in the job matrix because YAML would parse the unquoted version as 3.1, so I went ahead and added quotes around all entries in the `python-version` list for consistency. I can revert this if desired.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


droideck (approved):
LGTM!

PR #506, opened 2022-12-21
Fix CI images after ubuntu-latest update

droideck:
Keep 20.04 for py3.6 testing.
Replace deprecated enchant package with enchant-2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on: [push, pull_request]
 jobs:
   distros:
     name: "Ubuntu with Python ${{ matrix.python-version }}"
-    runs-on: "ubuntu-latest"
+    runs-on: "ubuntu-20.04"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


mistotebe:
Why freeze the Ubuntu version? Is python 3.6 not installable on ubuntu-latest at all? LGTM otherwise.

droideck:
> Why freeze the Ubuntu version? Is python 3.6 not installable on ubuntu-latest at all? LGTM otherwise.

Yep, Python 3.6 is no more on ubuntu-latest.
https://github.com/actions/setup-python/issues/544#issuecomment-1332535877

> Hello everyone. Python 3.6 [reached end-of-life on Dec 12, 2021](https://peps.python.org/pep-0494/#lifespan). Therefore, we won’t be supporting it with new versions of our hosted runners. If you need Python 3.6, it’s still available on ubuntu-20.04.
There are a couple reasons we stop supporting end-of-life versions. They no longer have maintenance support, so there’s no path forward for fixing bugs or security issues. Also, they tend to run into compat issues with dependencies on newer operating systems, which is a source of confusing bugs.

Issue #460, opened 2022-02-01
Tests/t_ldapobject.py fails on pypy3 CI run


droideck:
CI tests start to fail in Tests/t_ldapobject.py with the next error:

```
        self.server.slapadd(SCHEMA_TEMPLATE, ["-n0"])
        self.server.restart()
>       self.reset_connection()

# Or

      def tearDown(self):
  >       self._sock.close()
```

But always it points to:

```
    def _real_close(self, _ss=_socket.socket):
        # This function should not reference any globals. See issue #808164.
>       _ss.close(self)
E       OSError: [Errno 9] Bad file descriptor
```

Possibly, we're trying to close the socket that was already closed.
And, it's possibly a race condition because sometimes it happens and sometimes it's not (and only on pypy3 runs).

Further investigation is required.

droideck:
Fixed by https://github.com/python-ldap/python-ldap/commit/e5959b38902bbf1d68a3f16b04ddc210884b8d5e

droideck:
Reopening as the new failure has appeared in the same test as the tearDown issue was fixed:

```
=================================== FAILURES ===================================
___________ Test03_SimpleLDAPObjectWithFileno.test_reject_bytes_base ___________

self = <Tests.t_ldapobject.Test03_SimpleLDAPObjectWithFileno testMethod=test_reject_bytes_base>

    def setUp(self):
        try:
>           self._ldap_conn
E           AttributeError: 'Test03_SimpleLDAPObjectWithFileno' object has no attribute '_ldap_conn'

Tests/t_ldapobject.py:114: AttributeError

During handling of the above exception, another exception occurred:

self = <Tests.t_ldapobject.Test03_SimpleLDAPObjectWithFileno testMethod=test_reject_bytes_base>

    def setUp(self):
        try:
            self._ldap_conn
        except AttributeError:
            # open local LDAP connection
>           self._ldap_conn = self._open_ldap_conn(bytes_mode=False)

Tests/t_ldapobject.py:117: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
Tests/t_ldapobject.py:656: in _open_ldap_conn
    who=who, cred=cred, fileno=self._sock.fileno(), **kwargs
.tox/pypy3/site-packages/slapdtest/_slapdtest.py:598: in _open_ldap_conn
[...]

PR #507, opened 2023-01-13
fix(ReconnectLDAPObject): reconnect race condition

### LINKED ISSUE #253: SimpleLDAPObject has no attribute '_l'
Issue description:

After initializing and using simple_bind_s everything works great. But when it comes to unbind_s function it gives exception of AttributeError which is resulted to "SimpleLDAPObject has no attribute '_l'"

Steps to reproduce: 
initialize ldap, bind_s, then unbind will result to this error. 

Operating system: macOS 10.14
Python version: 3.6
python-ldap version: 3.1.0

pmhahn:
There is a race condition in ReconnectLDAPObject._apply_method_s() when
multiple threads use a shared ReconnectLDAPObject instance:

> Traceback (most recent call last):
...
>   File "ldap/ldapobject.py", line 1259, in search_ext_s
>     return self._apply_method_s(SimpleLDAPObject.search_ext_s,*args,**kwargs)
>   File "ldap/ldapobject.py", line 1201, in _apply_method_s
>     self.reconnect(self._uri,retry_max=self._retry_max,retry_delay=self._retry_delay)
>   File "ldap/ldapobject.py", line 1181, in reconnect
>     SimpleLDAPObject.unbind_s(self)
>   File "ldap/ldapobject.py", line 905, in unbind_s
>     return self.unbind_ext_s(None,None)
>   File "ldap/ldapobject.py", line 889, in unbind_ext_s
>     msgid = self.unbind_ext(serverctrls,clientctrls)
>   File "ldap/ldapobject.py", line 881, in unbind_ext
>     res = self._ldap_call(self._l.unbind_ext,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
>   File "ldap/ldapobject.py", line 352, in __getattr__
>     self.__class__.__name__,repr(name)
> AttributeError: ReconnectLDAPObject has no attribute '_l'

While the reconnect() is already protected by `self._reconnect_lock`,
there is a small windows between the `unbind_s()` and `reconnect()`
call, where CPython can switch threads: While the 1st threads is already
waiting inside `reconnect()` for its next try, the 2nd threads will call
`unbind_s()` outside that lock which will race with the 1st thread also
calling `unbind_s()`: Whoever comes first will do the `del self._l`
first, so the later threads will raise the above AttributeError.

Move calling `unbind_s()` inside the locked region so that `self._l` is
handled atomically.

Add a new parameter `force` to either forcefully close any previous
connection or keep re-use the previous connection if it still is
supposed to work.

PS: This does not fix #253 but addresses https://forge.univention.org/bugzilla/show_bug.cgi?id=43274

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -893,7 +893,7 @@ class ReconnectLDAPObject(SimpleLDAPObject):
     self._reconnect_lock = ldap.LDAPLock(desc='reconnect lock within %s' % (repr(self)))
     # XXX cannot pickle file, use default trace file
     self._trace_file = ldap._trace_file
-    self.reconnect(self._uri)
+    self.reconnect(self._uri,force=True)
 
   def _store_last_bind(self,_method,*args,**kwargs):
     self._last_bind = (_method,args,kwargs)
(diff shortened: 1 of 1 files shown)


pmhahn:
There seems to be a race-condition in running the tests in `pypy3`: I can "reproduce" the bug locally on my Debian-11-Bullseye machine, but mostly one test failed; which one is random. Sometimes I get a clean run, but most fail.
I know that `pypy3` has a different behavior on when objects are freed as this is unspecified and different Python implementations differ in that detail.

droideck:
Yeah, it's a known issue: https://github.com/python-ldap/python-ldap/issues/460. So your change won't 

Regarding your PR, it looks good to me! Thank you for your contribution!

I'll leave a bit of a time window for @mistotebe (or anybody else) to review, and we can merge it early next week.

droideck (approved):

PR #522, opened 2023-05-15
[python-ldap] Add type hints

Alphix:
With this patch applied, ``MYPYPATH=/Stubs mypy --strict Lib/`` passes without any warnings. ``tox`` also passes for everything from ``py37`` to ``py11`` (and also including ``doc`` and ``pypy3``).

I've tried my best to not make more modifications than strictly necessary. Most of the time, changes merely serve to make life easier for ``mypy`` (i.e. to enable type inference) or to fixup non-type aware code.

The patch might look dauntingly large, but once you cut away the trivial stuff (untyped function definitions which have been replaced with types ones, lots and lots of new import statements, etc) the diff is actually not that big.


mistotebe:
Hi, thanks for the PR!

Looks good, would you be able to split it into a few more commits each doing self-contained changes: any fixes, `ldap.pkginfo` changes, `compat.py` removal, etc. in any order that makes sense and then just adding the type hints only?

And just out of curiosity, is there a reason you chose to use `<type> | None` rather than `Optional[<type>]`?

tiran (dismissed):
Excellent work! A few comments and suggestions to push your PR over the finish line:

- [ ] Fix Python 3.6 compatibility issues. We still support old Python versions for $REASONS..., sorry.
- [ ] Please don't introduce yet another top-level module name. We like to move `ldapurl` and `ldif` into the `ldap` package eventually. I suggest `ldap.types` module.
- [ ] Don't use `from ldap_types import *`. Star-imports are frowned upon and make code harder to read.
- [ ] The PR is missing a PEP 561 type marker file. Add a `Lib/ldap/py.typed` file and add the file to `MANIFEST.in`
- [ ] Move `mypy.ini` content to either `pyproject.toml` or `setup.cfg`
- [ ] Extend `tox.ini` with a mypy check for every supported Python version
- [ ] Add `_ldap.pyi` to `MANIFEST.in`
- [ ] Would it make sense to move `_ldap.pyi` to `Lib/_ldap.pyi` and install it along the shared extension?

Python 3.6 fails because it does not support future annotations:
```
     File "/home/runner/work/python-ldap/python-ldap/.tox/c90-py36/lib/python3.6/site-packages/ldap/__init__.py", line 6
      from __future__ import annotations
                                       ^
  SyntaxError: future feature annotations is not defined
```

intgr:
Hi! Why has this this work stalled? @Alphix are you still interested in continuing this?

> Fix Python 3.6 compatibility issues. We still support old Python versions for $REASONS..., sorry.

What are those reasons? Many projects have already dropped 3.7.

Alphix:
> Hi! Why has this this work stalled? @Alphix are you still interested in continuing this?

@intgr - it's the usual reason, I need to find the time to take the comments into account and update the patch. Maybe next week, we'll see...

>> Fix Python 3.6 compatibility issues. We still support old Python versions for $REASONS..., sorry.
> 
> What are those reasons? Many projects have already dropped 3.7.

My problem is that ``tox py36`` fails even without the patch applied, haven't looked into that yet...

tiran:
> > Fix Python 3.6 compatibility issues. We still support old Python versions for $REASONS..., sorry.
> 
> What are those reasons? Many projects have already dropped 3.7.

Because we (downstream vendor) still support Python 3.6.

intgr:
> Because we (downstream vendor) still support Python 3.6.

My question was: why can't python-ldap drop support for Python 3.6 (and maybe 3.7)?

Alphix:
> > Because we (downstream vendor) still support Python 3.6.
> 
> My question was: why can't python-ldap drop support for Python 3.6 (and maybe 3.7)?

My guess would be RHEL?

Anyway, does ``tox py36`` work for anyone else? Without this patch applied? Because it doesn't for me....

tiran:
> > > Because we (downstream vendor) still support Python 3.6.
> > 
> > 
> > My question was: why can't python-ldap drop support for Python 3.6 (and maybe 3.7)?
> 
> My guess would be RHEL?
> 
> Anyway, does `tox py36` work for anyone else? Without this patch applied? Because it doesn't for me....

`tox -e py36` is passing on my Fedora machine. It's also working fine on our CI, which runs the tests both on Ubuntu and Fedora.

Alphix:
> Looks good, would you be able to split it into a few more commits each doing self-contained changes: any fixes, `ldap.pkginfo` changes, `compat.py` removal, etc. in any order that makes sense and then just adding the type hints only?

Sorry, but I think that would just be way too much work at the moment. The changes to the code mostly go hand-in-hand with the typing annotations (and in most cases are just there to make the types more explicit to `mypy`).

> And just out of curiosity, is there a reason you chose to use `<type> | None` rather than `Optional[<type>]`?

That was an oversight, didn't realise it was relevant for backwards compatibility, I've fixed that in the second version.

Alphix:
> Excellent work! A few comments and suggestions to push your PR over the finish line:
> 
> * [x]  Fix Python 3.6 compatibility issues. We still support old Python versions for $REASONS..., sorry.
> * [x]  Please don't introduce yet another top-level module name. We like to move `ldapurl` and `ldif` into the `ldap` package eventually. I suggest `ldap.types` module.
> * [x]  Don't use `from ldap_types import *`. Star-imports are frowned upon and make code harder to read.
> * [x]  The PR is missing a PEP 561 type marker file. Add a `Lib/ldap/py.typed` file and add the file to `MANIFEST.in`
> * [x]  Move `mypy.ini` content to either `pyproject.toml` or `setup.cfg`
> * [x]  Extend `tox.ini` with a mypy check for every supported Python version
> * [x]  Add `_ldap.pyi` to `MANIFEST.in`
> * [x]  Would it make sense to move `_ldap.pyi` to `Lib/_ldap.pyi` and install it along the shared extension?

@tiran : I think I've addressed all the points you raised (and the CI seems happy). The only thing I can think of is that `mypy` is not enabled for the `py312` environment right now because it needs different compile flags. Not sure what the best way to handle that in `tox.ini` would be....but perhaps it could be addressed in a separate patch?

Also, I should add that this is just a starting point for type annotations. There are still a bunch of uses of `Any` and `type: ignore` in the code. I already have another branch which gradually fixes up those and makes the type checking stricter (but that's for later).

Alphix:
> > Looks good, would you be able to split it into a few more commits each doing self-contained changes: any fixes, `ldap.pkginfo` changes, `compat.py` removal, etc. in any order that makes sense and then just adding the type hints only?
> 
> Sorry, but I think that would just be way too much work at the moment. The changes to the code mostly go hand-in-hand with the typing annotations (and in most cases are just there to make the types more explicit to `mypy`).

Scratch that, I am working on splitting the PR into several preparatory patches, one big type hints patch, and one final patch integrating type checking (and learning a lot about `git add -p` in the process).

So you might want to hold off on the review for now...

Alphix:
> Looks good, would you be able to split it into a few more commits each doing self-contained changes: any fixes, `ldap.pkginfo` changes, `compat.py` removal, etc. in any order that makes sense and then just adding the type hints only?

Ok, done. I've split the patch into a bunch of commits doing a lot of prep work, then one big patch adding the type hints only and finally one patch integrating `mypy` into the build machinery.

Lib/ldap/controls/libldap.py
```diff
@@ -26,14 +28,18 @@ class AssertionControl(RequestControl):
...
     self.filterstr = filterstr
 
-  def encodeControlValue(self):
-    return _ldap.encode_assertion_control(self.filterstr)
+  def encodeControlValue(self) -> bytes:
+    return _ldap.encode_assertion_control(self.filterstr)  # type: ignore
 
-KNOWN_RESPONSE_CONTROLS[ldap.CONTROL_ASSERT] = AssertionControl
+# FIXME: This is a request control though?
+#KNOWN_RESPONSE_CONTROLS[ldap.CONTROL_ASSERT] = AssertionControl
```
intgr, line 42: FIXME should be fixed. Just pointing it out, not sure what to do here.

Lib/ldap/types.py
```diff
@@ -0,0 +1,58 @@
...
+    'LDAPEntryDict',
+    'LDAPControl',
+    'LDAPControls',
+    'LDAPSearchResult',
+]
+
+LDAPModListAddEntry: TypeAlias = "Tuple[str, List[bytes]]"
+"""The type of an addition entry in a modlist."""
+
+LDAPModListModifyEntry: TypeAlias = "Tuple[int, str, Optional[List[bytes]]]"
```
intgr, line 33: `modify_s()` also supports bytes in addition to `list[bytes] | None`

```suggestion
LDAPModListModifyEntry: TypeAlias = "Tuple[int, str, Union[List[bytes], bytes, None]]"
```

https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#ldap.LDAPObject.modify

> mod_vals is either a string value or a list of string values to add, delete or replace respectively.

intgr: Also in practice it accepts `int`, my code has been using that and it works.

But it's not documented. Should we err on the side of strictness and not allow `int` here?

Lib/_ldap.pyi
```diff
@@ -0,0 +1,406 @@
...
+URL_ERR_MEM: int
+VENDOR_VERSION: int
+VERSION: int
+VERSION1: int
+VERSION2: int
+VERSION3: int
+VERSION_MAX: int
+VERSION_MIN: int
+
+class ADMINLIMIT_EXCEEDED(ldap.LDAPError):
```
intgr, line 187: Better move `LDAPError` definition to the top and avoid circular import...

```suggestion
class ADMINLIMIT_EXCEEDED(LDAPError):
```

Lib/_ldap.pyi
```diff
@@ -0,0 +1,406 @@
...
+class INVALID_DN_SYNTAX(ldap.LDAPError):
+    errnum: ClassVar[int] = ...
+
+class INVALID_SYNTAX(ldap.LDAPError):
+    errnum: ClassVar[int] = ...
+
+class IS_LEAF(ldap.LDAPError):
+    errnum: ClassVar[int] = ...
+
+class LDAPError(Exception): ...
```
intgr, line 271: Mypy does not consider these error classes and constants to be re-exported from the `ldap` module.

So when I use `import ldap; ldap.SERVER_DOWN`, it causes "error: Module has no attribute "SERVER_DOWN" [attr-defined]". Same with constants such as `SCOPE_SUBTREE`, `MOD_REPLACE` etc.

`Lib/ldap/__init__.py` contains `from _ldap import *`. I'm a bit confused though, https://typing.readthedocs.io/en/latest/source/stubs.html#imports lists `from Y import *` as being a re-export. Not sure what is happening.

setup.py
```diff
@@ -157,6 +157,7 @@ class OpenLDAP2:
   install_requires=[
     'pyasn1 >= 0.3.7',
     'pyasn1_modules >= 0.1.5',
+    'typing_extensions',
```
intgr, line 160: Should probably have a versioned dependency here. This is imported runtime and a too old version will cause `ImportError`.

Have a look in which version the symbols were added that you're using: https://github.com/python/typing_extensions/blob/main/CHANGELOG.md

intgr (commented):
Awesome, thanks for your work. I tried it out in my project and have some feedback.

Lib/_ldap.pyi
Alphix: > ldap.SERVER_DOWN

Actually, I realised that adding `py.typed` and `_ldap.pyi` to `MANIFEST.in` wasn't sufficient (and if you check the `python_ldap-*` dir in `site-packages` (presumably in your virtual environment), the files aren't there.

I fixed that locally by adding `extra_files = .:Lib/_ldap.pyi` to the `[_ldap]` section of `setup.cfg` and adding `package_data = {'ldap': ['py.typed'],},` towards the bottom of `setup.py`. The files now get installed, but `mypy` still isn't able to find the `_ldap.pyi` stub... 😕

Lib/_ldap.pyi
Alphix: I'd rather leave things like they are in order to keep the `.pyi` file as similar to the original output of `stubgen` as possible...correct me if I'm wrong, but circular imports shouldn't be much of a concern with `.pyi` files?

Lib/ldap/types.py
Alphix: I think the type hints can be stricter than what the code actually accepts...they're supposed to document the "right" way of doing things after all.

Lib/ldap/controls/libldap.py
Alphix: Yeah, I'm not sure that I'm right here, so I kind of left it as a marker for the maintainers to have a look. I guess the "fix" would be just deleting the lines, but not sure

Lib/ldap/types.py
Alphix: And good catch WRT `mod_vals`...I also noticed that the `timeout` parameter type (in various functions in `Lib/ldap/ldapobject.py`) is too strict (it accepts `int | float`)

Lib/_ldap.pyi
Alphix: I already tried that (and also tried creating an empty top-level `__init__.py` file)...still no dice

intgr: Here some solutions are discussed:

* https://github.com/python/typing/issues/1333

One approach is to convert the `_ldap` module to a package first.

But I wonder if including an `_ldap-stubs` directory with `__init__.pyi` would also solve the issue.

Lib/_ldap.pyi
intgr: > correct me if I'm wrong, but circular imports shouldn't be much of a concern with .pyi files?

That's true. But I think following `stubgen` output should be a non-goal. stubgen is something a project should only do once, and for the rest of the lifetime, `.pyi` files will be for humans to read and edit. So I would emphasize readability over conformance with stubgen output.

If you want to ensure the `.pyi` doesn't get outdated compared to the implementation, it's much better to run `stubtest` in CI.

Lib/_ldap.pyi
Alphix: > Here some solutions are discussed, but they're not very nice:
> 
> * [Type stubs for single-file top-level modules python/typing#1333](https://github.com/python/typing/issues/1333)

Hm, good link....but no real solutions IMHO.

> One approach is to convert the `_ldap` module to a package first.
> 
> But I wonder if including an `_ldap-stubs` directory with `__init__.pyi` would also solve the issue.

I tried manually moving `_ldap.*` into the `.../lib/python3.x/site_packages/python_ldap-*/ldap` directory and changing a few imports, and that actually seems to work. Would mean that `_ldap` is instead `ldap._ldap`, but given that it's an internal implementation detail, that might actually be ok...I'll work on a patch to see what it might look like...

Alphix: > > Here some solutions are discussed, but they're not very nice:
> > 
> > * [Type stubs for single-file top-level modules python/typing#1333](https://github.com/python/typing/issues/1333)
> 
> Hm, good link....but no real solutions IMHO.
> 
> > One approach is to convert the `_ldap` module to a package first.
> > But I wonder if including an `_ldap-stubs` directory with `__init__.pyi` would also solve the issue.
> 
> I tried manually moving `_ldap.*` into the `.../lib/python3.x/site_packages/python_ldap-*/ldap` directory and changing a few imports, and that actually seems to work. Would mean that `_ldap` is instead `ldap._ldap`, but given that it's an internal implementation detail, that might actually be ok...I'll work on a patch to see what it might look like...

Ok, I've managed to reflect that in the build system as well. I'll go with that solution for now (or until someone smarter than me comes up with a better solution).

Alphix:
@tiran gentle ping?

macserv:
Hey, @Alphix, can you update this to resolve conflicts, and then page @droideck, @mistotebe, and @tiran for review?

Thanks so much... this is a tremendous contribution that shouldn't be left to bit-rot.

Alphix:
> Hey, @Alphix, can you update this to resolve conflicts, and then page @droideck, @mistotebe, and @tiran for review?

@macserv Eh, I'm a volunteer, using my free time for this. If *you* want to push this forward, I'd suggest *you* try to get in touch with the maintainers...once there's some sign of interest from their side, I'd be happy to rebase and resolve conflicts...

spaceone:
> Most controversial change here is the move of `_ldap` to `ldap._ldap` (which is why that's a separate patch)

Why is this necessary?
The commit message says:

> This is necessary in order for type checking due to:
> https://github.com/python/typing/issues/1333
> 
> Given that _ldap is an internal module, this change is hopefully ok.

That upstream issue is still open.

Lib/ldap/controls/libldap.py
```diff
@@ -47,14 +53,19 @@ class MatchedValuesControl(RequestControl):
...
     self.filterstr = filterstr
 
-  def encodeControlValue(self):
-    return _ldap.encode_valuesreturnfilter_control(self.filterstr)
+  def encodeControlValue(self) -> bytes:
+    return _ldap.encode_valuesreturnfilter_control(self.filterstr)  # type: ignore
 
-KNOWN_RESPONSE_CONTROLS[ldap.CONTROL_VALUESRETURNFILTER] = MatchedValuesControl
+# FIXME: This is a request control though?
+#KNOWN_RESPONSE_CONTROLS[ldap.CONTROL_VALUESRETURNFILTER] = MatchedValuesControl
```
spaceone, line 68: @Alphix yes, `MatchedValuesControl` is a request control.

Lib/ldap/controls/libldap.py
Alphix: Yeah, that can be fixed in a subsequent patch though...

Alphix:
> @Alphix Sorry for the late reply! Maintainer is here! Thank you for your contribution!
> 
> If you are still around, could you please rebase it?

@droideck: Ok, done...please have a look

mistotebe:
Hi @Alphix , I broadly support the changes, sorry that it's languished in review for a while. I'll have a more detailed look with other's comments in mind. In the meantime, could you generate the `.pyi` file with `Lib/ldap/constants.py` so it stays up to date if our native interface changes?

Lib/ldap/controls/libldap.py
Alphix: It's fixed in the new version I just pushed, deleted instead of FIXME

Lib/ldap/controls/libldap.py
Alphix: It's fixed in the new version I just pushed, deleted instead of FIXME

Alphix:
> Hi @Alphix , I broadly support the changes, sorry that it's languished in review for a while. I'll have a more detailed look with other's comments in mind. In the meantime, could you generate the `.pyi` file with `Lib/ldap/constants.py` so it stays up to date if our native interface changes?

@mistotebe Ok, I've pushed an updated version. I hope I understood what you meant with "generate the `.pyi` file with `Lib/ldap/constants.py`" correctly. The last 5 commits in this PR is the new stuff since the last version. All tests pass. I hope this can be committed soon, because it's not a very pleasant PR to rebase (which is why I used claude to help me with some of the more mechanical refactoring).

@droideck @tiran Please have a look

mistotebe:
@Alphix I've updated your PR against master, can you double check you're happy with what I've done? If you can look over @tiran's comments and make sure you addressed all of them (it looks to me like you have) let me know and I am happy to merge afterwards.

spaceone:
I am a little bit unhappy about moving `_ldap` to `ldap._ldap` and that is also kind of an API change - but yes, it's a private module.
For some things I import `_ldap`.

Is this really necessary? for what? a mypy workaround?
Nowadays there is `ty`, which is also a very good type checker. Is that affected as well?

Alphix:
> Is this really necessary? for what? a mypy workaround?

I must admit that my memory is a bit (very) foggy given the age of the PR, but I think you already provided the same question (https://github.com/python-ldap/python-ldap/pull/522#issuecomment-3162916268) and I replied to it (https://github.com/python-ldap/python-ldap/pull/522#issuecomment-3190009202).

> Nowadays there is `ty`, which is also a very good type checker. Is that affected as well?

No clue, feel free to test it yourself and share your findings?

Alphix:
> @Alphix I've updated your PR against master, can you double check you're happy with what I've done?

@mistotebe I've had a quick look, and it seems ok to me.

> If you can look over @tiran's comments and make sure you addressed all of them (it looks to me like you have) let me know and I am happy to merge afterwards.

Yeah, as far as I know I addressed them back in 2024, and I don't think @tiran has said anything to indicate that there's lingering issues...

mistotebe:
> I am a little bit unhappy about moving `_ldap` to `ldap._ldap` and that is also kind of an API change - but yes, it's a private module. For some things I import `_ldap`.

Hmm, can you share what's in `_ldap` that you prefer over the `ldap` interfaces? Are you accessing `_ldap.LDAP`'s methods directly?

Overall we'd like to move most modules (ldif, ldapurl etc.) under ldap, because the pypi namespace is polluted with unrelated modules and this is causing real (discoverability) issues: e.g. `pypi install ldif` might not give you what you wanted.

spaceone:
We maintain a linux distribution based on Debian 12 and Debian 13 (and earlier releases).
We are heavy users of python-ldap and rely on it's functionality even during package upgrades or Debian release upgrades.
There is a time-frame in which the Debian packages are half-unpacked and a import of Python-ldap not available - e.g. in asynchronous components or scripts running in Debian pre-installation maintainer scripts.
When moving library imports, we observed race-conditions where the import gets non-functional for a short amount of time.
We didn't have this kind of problems with python-ldap yet, but with other not so critical libraries.

Also we are very happy at the moment, that we could drop all self-maintained patches for python-ldap and use the upstream packages from Debian again.
Especially because of securty monitoring - each forked package adds a burden.
This will change with https://github.com/python-ldap/python-ldap/pull/618 again :-(

And we really need the https://github.com/python-ldap/python-ldap/pull/466 for performance, security and correctness reasons.
As we are going to fork python-ldap again, we will include this now manually again.

Semantic versioning wants for breaking changes a major version update, for features a minor version increase and for fixes a patchlevel increase.

So, I would consider the VLV thing a bugfix with a patchlevel increase.
The DN string is a feature, which would be a minor update.
And the moving of these libraries is a breaking change - major update.

I am not sure, if Debian would just integrate the breaking change - if it looks like a non-major-change version number - into a point release. That would be much effort for us.

So, I would be happy, if such library moves would still provide an backwards compatible alias:
`_ldap.py` → `from ldap._ldap import *`
`ldif.py` → `from ldap.ldif import *`

python-ldap doesn't seem to follow semantic versioning.
(So I also don't see why the DN feature - which is a completely backwards compatible feature/not changing any semantics - cannot be taken in one of the next releases.)

mistotebe:
@Alphix we're almost there, I don't think we can use `ldap.types` because it actively conflicts with Python's `types` module, we might rename it to `_types` and maybe reexport the TypeAliases from `ldap` proper?

@spaceone OK, we should be able to reexport in `_ldap`. `ldif`, `ldapurl` etc. would probably come later but the same reexport and a deprecation with a planned removal in 5 (and a Warning in 4) might be the way to go (what do you think @droideck?). All the other issues you mentioned should now be sorted in the relevant branches, feel free to comment there if you disagree.

I'd also be minded to enabling some baseline type-checking for the CI, but no need to do it here, happy to get it in myself later, just like making sure that `Lib/ldap/constants.py` is clean on the tree.

mistotebe:
I'm also very happy to push ahead with doing the above next week myself, barring any objections of course.

mistotebe:
Hmm, mypy can't figure out that respoid must be `None` in 3.9 (only), not sure why. Don't know whether it's worth adding the `assert isinstance(respoid, None)` if we're going to drop 3.9,3.10 soon because of the Stable ABI work

Lib/ldap/ldapobject.py
```diff
@@ -449,24 +644,53 @@ def modrdn(self,dn,newrdn,delold=1):
...
+    newpw: str,
+    serverctrls: List[RequestControl] | None = None,
+    clientctrls: List[RequestControl] | None = None,
+    extract_newpw: bool = False,
+  ) -> Tuple[None, bytes | PasswordModifyResponse]:
     msgid = self.passwd(user, oldpw, newpw, serverctrls, clientctrls)
     respoid, respvalue = self.extop_result(msgid, all=1, timeout=self.timeout)
 
     if respoid != PasswordModifyResponse.responseName:
       raise ldap.PROTOCOL_ERROR("Unexpected OID %s in extended response!" % respoid)
```
droideck, line 669: @mistotebe about the 3.9 mypy issue... Yeah, looks like the tox env installs mypy unpinned, so 3.9 gets 1.19.1 (the last release that runs on 3.9) while the other jobs get 2.3.1, and only 1.19.1 can't tell that respoid is None after this raise. IIUC, `isinstance(respoid, None)` won't work since isinstance needs a type, but "assert respoid is None" right below this line should fix it for both versions.
It's one line, so I'd add it even if 3.9 goes away soon.

Lib/ldap/resiter.py
```diff
@@ -3,16 +3,26 @@
...
+import ldap
 
 from ldap.pkginfo import __version__, __author__, __license__
 
+from ldap.controls import ResponseControl
 
-class ResultProcessor:
+from typing import Any, List, Tuple, Iterator
+
+class ResultProcessor(ldap.ldapobject.LDAPObject):
```
droideck: As ResultProcessor is a mixin, making it a subclass of LDAPObject breaks the documented usage: class `MyLDAPObject(LDAPObject, ResultProcessor)` now fails with an MRO error, because LDAPObject comes before its own subclass. The docs and three demos use exactly that order.
For mixins, IIRC, the usual trick is to inherit from LDAPObject under TYPE_CHECKING and from object otherwise.

Lib/ldap/controls/openldap.py
```diff
@@ -42,15 +43,21 @@ def decodeControlValue(self,encodedControlValue):
 ldap.controls.KNOWN_RESPONSE_CONTROLS[SearchNoOpControl.controlType] = SearchNoOpControl
 
 
-class SearchNoOpMixIn:
+class SearchNoOpMixIn(ldap.ldapobject.SimpleLDAPObject):
```
droideck: Same problem as `ResultProcessor`: class MyLDAPObject(LDAPObject, SearchNoOpMixIn) now fails with an MRO error. Same fix suggestion here.

Lib/ldap/schema/subentry.py
```diff
@@ -190,11 +266,12 @@ def tree(self,schema_element_class,schema_element_filters=None):
         # Ignore schema elements not matching schema_element_class.
         # This helps with falsely assigned OIDs.
         continue
+      # FIXME: This assertion is superfluous?
       assert se_obj.__class__==schema_element_class, \
         "Schema element referenced by {} must be of class {} but was {}".format(
           se_oid,schema_element_class.__name__,se_obj.__class__
         )
-      for s in se_obj.sup or ('_',):
+      for s in getattr(se_obj, "sup", ['_']):
```
droideck: This breaks tree(). The old code fell back to `'_'` when sup was empty. getattr only falls back when sup is missing, and it never is. So root elements are no longer listed under `'_'` and the tree can't be walked from the top (Demo/schema_tree.py prints nothing). Maybe `getattr(se_obj, "sup", ()) or ('_',)` to keep the fallback for both cases?

Lib/ldap/schema/models.py
```diff
@@ -639,24 +701,31 @@ def __str__(self):
...
-  def __init__(self,schema,dn,entry):
-    self._keytuple2attrtype = {}
-    self._attrtype2keytuple = {}
+  def __init__(self, schema: "ldap.schema.subentry.SubSchema", dn: str, entry: LDAPEntryDict) -> None:
+    self._keytuple2attrtype: Dict[Tuple[str, ...], str] = {}
+    self._attrtype2keytuple: Dict[str, Tuple[str, ...]] = {}
+    # This class wants to act like it's a string-keyed dict, but under the
+    # hood it uses the tuple of OID and sub-types of an attribute type
+    # as the key, so we can't use the self.data dict and stay type-safe.
+    self._data: Dict[Tuple[str, ...], List[bytes]] = {}
```
droideck: Storing values in self._data leaves the UserDict len and iteration looking at the empty self.data. So len(entry) is 0, iterating gives nothing and bool(entry) is False, while keys() and items() still work because they're overridden below. Needs matching overrides for length and iteration, or going back to self.data with a type: ignore.

Lib/ldap/controls/psearch.py
```diff
@@ -57,23 +61,28 @@ class PersistentSearchControlValue(univ.Sequence):
...
-      for ct in self.changeTypes:
-        changeTypes_int = changeTypes_int|CHANGE_TYPES_INT.get(ct,ct)
-      self.changeTypes = changeTypes_int
+    self.changeTypes = changeTypes or CHANGE_TYPES_INT.keys()
+
+  def encodeControlValue(self) -> bytes:
+    # Assume a sequence type of names of integers to be OR-ed
+    changeTypes_int = 0
+    for ct in self.changeTypes:
+      changeTypes_int |= CHANGE_TYPES_INT.get(ct, 0)
```
droideck: On main changeTypes could be an int bitmask or a list of ints. Now an int raises TypeError in the loop above, and a list of ints silently encodes 0 because of the .get(ct, 0). I'm fine with the hint staying List[str], as said earlier hints can be stricter than the code, but the runtime should keep working. Should we handle int separately and use .get(ct, ct) like main did?

Lib/ldap/controls/simple.py
```diff
@@ -105,8 +115,8 @@ class ProxyAuthzControl(RequestControl):
     on behalf which the server should process the request
   """
 
-  def __init__(self,criticality,authzId):
-    RequestControl.__init__(self,ldap.CONTROL_PROXY_AUTHZ,criticality,authzId)
+  def __init__(self, criticality: bool, authzId: str) -> None:
+    RequestControl.__init__(self,ldap.CONTROL_PROXY_AUTHZ,criticality,authzId.encode('utf-8'))
```
droideck: ProxyAuthzControl(True, b'dn:...') worked on main and now fails on the .encode. We can do the same as in psearch suggestion: keep the str hint, but only encode when it actually is a str.

Lib/ldap/controls/simple.py
```diff
@@ -141,6 +151,7 @@ class GetEffectiveRightsControl(RequestControl):
   """
   Get Effective Rights Control
   """
+  controlType = '1.3.6.1.4.1.42.2.27.9.5.2'
 
-  def __init__(self,criticality,authzId=None):
-    RequestControl.__init__(self,'1.3.6.1.4.1.42.2.27.9.5.2',criticality,authzId)
+  def __init__(self, criticality: bool, authzId: str) -> None:
```
droideck: authzId lost its None default, so GetEffectiveRightsControl(True) no longer works. We can "authzId: str | None = None" and encode only when set. WDYT?

Lib/ldap/controls/pagedresults.py
```diff
@@ -18,29 +18,43 @@
...
+    size: int = 10,
+    cookie: Union[str, bytes] = '',
+  ) -> None:
     self.criticality = criticality
     self.size = size
-    self.cookie = cookie or ''
 
-  def encodeControlValue(self):
+    if cookie is None:
+      cookie = b''
```
droideck: This sets the local variable, not self.cookie, so cookie=None now fails at encode time. It worked on main...

Lib/ldap/ldapobject.py
```diff
@@ -781,7 +1109,17 @@ def read_subschemasubentry_s(self,subschemasubentry_dn,attrs=None):
     else:
       return subschemasubentry
 
-  def find_unique_entry(self,base,scope=ldap.SCOPE_SUBTREE,filterstr=None,attrlist=None,attrsonly=0,serverctrls=None,clientctrls=None,timeout=-1):
+  def find_unique_entry(
+    self,
+    base: str,
+    scope: int,
```
droideck: scope lost its SCOPE_SUBTREE default, so find_unique_entry(base, filterstr=...) calls break.

Lib/ldif.py
```diff
@@ -202,10 +224,14 @@ def unparse(self,dn,record):
...
     # Count records written
     self.records_written = self.records_written+1
-    return # unparse()
 
 
-def CreateLDIF(dn,record,base64_attrs=None,cols=76):
+def CreateLDIF(
+    dn: str,
+    record: Union[LDAPEntryDict, LDAPModList],
+    base64_attrs: List[str],
```
droideck: base64_attrs used to default to None. Making it mandatory breaks CreateLDIF(dn, record)...

Lib/ldap/schema/__init__.py
```diff
@@ -4,7 +4,12 @@
...
 """
 
-from ldap import __version__
+from ldap.pkginfo import __version__
 
 from ldap.schema.subentry import SubSchema,SCHEMA_ATTRS,SCHEMA_CLASS_MAPPING,SCHEMA_ATTR_MAPPING,urlfetch
 from ldap.schema.models import *
+
+
+__all__ = [
```
droideck: With only SCHEMA_ATTRS in the export list, a star import from ldap.schema no longer gives you SubSchema, ObjectClass, AttributeType and so on. So I think we should either list them all or drop the list.

pyproject.toml
```diff
@@ -48,10 +46,10 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Topic :: System :: Systems Administration :: Authentication/Directory :: LDAP",
 ]
-
 dependencies = [
     "pyasn1>=0.3.7",
     "pyasn1_modules>=0.1.5",
+    "typing_extensions>=0.4.1",
```
droideck: typing_extensions 0.4.1 doesn't exist... We use TypeAlias and Self, and Self only appeared in 4.0.0, so that would be the real minimum (this is what @intgr asked for back in 2024).
IMO, even better is to drop the runtime dependency. All four uses are annotation only, so the imports can go under "if TYPE_CHECKING:" (tokenizer.py also needs the future annotations import). Debian 11 and Ubuntu 22.04 ship versions without Self, so this matters for distro installs.

tox.ini
```diff
@@ -91,6 +92,15 @@ basepython = pypy3
 deps = pytest
 commands = {envpython} -m pytest {posargs}
 
+[testenv:mypy]
+basepython = python3
+deps =
+    mypy
```
droideck: Please pin mypy and types-pyasn1, this is what made the 3.9 job red.
Unpinned, 3.9 gets mypy 1.19.1 with older stubs and the other jobs get 2.3.1 with newer stubs, and they disagree: the newer pair flags the five type: ignore[assignment] comments as unused, the older pair needs them. I think it's the only reason the warn_unused_ignores overrides in setup.cfg exist. With pinned versions (1.19.1 and types-pyasn1 0.6.0.20250914 while we support 3.9) the overrides can go, and once 3.9 and 3.10 are dropped for the stable ABI work the pin just moves to 2.x.

Lib/_ldap.py
```diff
@@ -0,0 +1 @@
+from ldap._ldap import *
```
droideck, line 9: The shim covers @spaceone's case: on the built wheel _ldap.LDAPError is ldap.LDAPError and the constants resolve. It doesn't forward the dunder version, author and license attributes, nor _VENDOR_VERSION_RUNTIME though (main's _ldap had all four), so those should be added explicitly. Plus a DeprecationWarning here if the plan is warn in 4, drop in 5.

droideck (changes requested):
Thanks for pushing this through the rebase @mistotebe!
And thanks @Alphix for sticking with it for so long:D

Okay, the typing work itself is in a really good shape
I think all @tiran's points from 2023 are covered, so his old "changes requested" can be dismissed at merge time.

Re-exporting the aliases from ldap proper sounds fine, but the LDAPControl alias in _types.py should get another name first, IMO, since ldap.controls.LDAPControl already exists. And yes to the _ldap shim with a warning in 4 and removal in 5, that also answers @spaceone's concern, I think. The shim is there already but incomplete (see the inline comment below)

Besides the comments, one thing to decide... There's no CHANGES entry yet, and the PR removes ldap.compat and ldap.DummyLock and moves _ldap.
@spaceone's point about semantic versioning applies here. To me this reads like a 4.0 change, unless we split the removals out or keep the shim silent for 3.5.0. Either way, I'd decided it explicitly

Some other intentional fixes also deserve a CHANGES line, IMO.
- passwordExpired can finally be True,
- attribute_types() now takes bytes objectClass values instead of str,
- the wrong KNOWN_RESPONSE_CONTROLS registrations are gone,
- the _s methods return None as documented,
- the default syncrepl cookie is '' instead of None,
- and ldap.extop no longer star-exports ExtendedRequest and ExtendedResponse.

What do you think?

Lib/_ldap.py
mistotebe: I'll update the shim in the next revision, but for the Warning, we probably need a separate issue since that has to land well after this PR.

Lib/ldap/schema/__init__.py
mistotebe: Not sure we know what to list at this point, especially given we import all from `ldap.schema.models`. If we want zero regressions for people who might be importing anything and everything we have in scope there, probably have to disable this mypy warning altogether? Or we could draw a line somewhere but not sure where yet.

Lib/ldap/ldapobject.py
mistotebe: I'm actually considering marking the module 3.11 right now. We know it's going to happen as the next thing after this and before 3.5 so we don't have to pretend to support something we don't intend to?

droideck: If possible, I'd rather not go 3.11 right now, it makes life harder for some distros. RHEL 9 for example is on 3.9 and I'd like to get 3.5 in there at some point. I'd drop 3.9/3.10 in 4.0 together with the stable ABI work.
IMO, for this PR I'd just pin mypy and types-pyasn1 in tox (or run the mypy env on a single interpreter), add the one line assert plus put typing_extensions under TYPE_CHECKING.

Also curious what in the stable ABI work actually needs 3.11... abi3 wheels can target 3.9 too, IIRC.

Lib/ldap/schema/__init__.py
droideck: Oh, looked at why the list is there... IIUC, it's only because ldapobject.py line 43 does "from ldap.schema import SCHEMA_ATTRS" and mypy strict doesn't like implicit re-exports. If that line imports from ldap.schema.subentry instead, I think, we can drop the list completely.

mistotebe:
I've made some updates, I **think** I covered the bits marked resolved, keeping them (mostly, there was a lot of history rewriting) as fixup commits in the place where they would fold in.

Lib/ldap/schema/__init__.py
mistotebe: Ok, done this and yes, mypy is happy.

Lib/ldap/ldapobject.py
mistotebe: Stable ABI + per module state can't really be done at the same time until 3.11, supporting 3.9 from that codebase would be even bigger undertaking...

mistotebe:
> Besides the comments, one thing to decide... There's no CHANGES entry yet, and the PR removes
> ldap.compat

I've tentatively reinstated that module. Wasn't hard to make mypy happy so not going to fight this.

> ldap.DummyLock

I would push back on this one, do we really care? This has no use since 3.7 and noone should be interested either.

> and moves _ldap. @spaceone's point about semantic versioning applies here. To me this reads like a 4.0 change, unless we split the removals out or keep the shim silent for 3.5.0. Either way, I'd decide it explicitly

Should be perfectly backwards compatible now I hope

> Some other intentional fixes also deserve a CHANGES line, IMO.
> 
>     * passwordExpired can finally be True,
>     * the wrong KNOWN_RESPONSE_CONTROLS registrations are gone,

Sure

>     * attribute_types() now takes bytes objectClass values instead of str,

I'm not sure what you mean.

>     * the _s methods return None as documented,
>     * the default syncrepl cookie is '' instead of None,

Sound like something that should not change, can you show me where?

>     * and ldap.extop no longer star-exports ExtendedRequest and ExtendedResponse.

Star? (I mean I've now reexported those two because we used to so maybe there's no such difference against `main` anymore)

> What do you think?

Thanks for the review, rewriting the patchset so many times I think I'm incapable of actually looking at the diffs with fresh eyes so this is much appreciated.

Lib/ldap/controls/__init__.py
```diff
@@ -8,16 +8,19 @@
 Each class provides support for a certain control.
 """
 
+from __future__ import annotations
 from ldap.pkginfo import __version__
 
-import _ldap
+import ldap._ldap as _ldap
```
spaceone:
```suggestion
from ldap import _ldap
```

(here and everywhere else)

Lib/ldap/controls/__init__.py
```diff
@@ -8,16 +8,19 @@
...
 assert _ldap.__version__==__version__, \
        ImportError(f'ldap {__version__} and _ldap {_ldap.__version__} version mismatch!')
 
 import ldap
 
 from pyasn1.error import PyAsn1Error
 
+from typing import Dict, List, Tuple, Type
```
spaceone: if only py3.11 (or even 3.9) is supported you can use the original types

Lib/_ldap.py
mistotebe: Created #640

droideck:
> > ldap.DummyLock
> 
> I would push back on this one, do we really care? This has no use since 3.7 and noone should be interested either.

Fair enough, it's been dead since 3.7 and nobody can have a use for it. Fine with dropping it.

> > ```
> > * attribute_types() now takes bytes objectClass values instead of str,
> > ```
> 
> I'm not sure what you mean.
> 

Sorry, that was too terse..:D
On main, Entry.attribute_types() worked when the objectClass values were str and blew up on bytes. Now it decodes them (models.py:795 - `ocs = [oc.decode("utf-8") for oc in bin_ocs]`), so bytes work, which is what search results gives you, but str values fail with AttributeError on .decode. IMO, decode only when it's bytes and both keep working.

> > ```
> > * the _s methods return None as documented,
> > * the default syncrepl cookie is '' instead of None,
> > ```
> 
> Sound like something that should not change, can you show me where?

syncrepl.py:527, syncrepl_get_cookie() returns `''` where main returned `None`. syncrepl_search() passes it straight into SyncRequestControl (:406-408) and encodeControlValue adds the syncCookie element whenever the cookie is not None (:107). So a consumer that doesn't override the hook now sends an empty cookie instead of none. I'd keep returning None there (-> str | None).

Lib/ldap/ldapobject.py
droideck: I've started looking into #641 and I'm fine with stable ABI landing in 3.5, just in case, that's not what I'm pushing back on...

I think #641 doesn't need the 3.11 floor for it. IIUC, the only thing in there that doesn't compile before 3.11 is `PyType_GetModuleByDef`. Since _ldap.LDAP can't be subclassed, `PyType_GetModule(Py_TYPE(self))` does the same job and has existed since 3.9, so 3.9/3.10 keep building the normal version-specific extension like they do today.
The abi3 wheel is unaffected, it is still built on 3.13+ only. The #ifdef guards and the l_ldap_new fallback already in the PR are the 3.9 bits, so IMO it's not a big undertaking.

So the proposal is this: 3.5 = #522 + #641 with the floor staying at >=3.9, CHANGES may say 3.5 is the last series for 3.9/3.10, and 4.0 drops them. We get stable ABI in 3.5, and we keep the ability to backport 3.5 to RHEL 9 and Ubuntu 22.04 (BTW, 3.10 is still ~28% of our pip downloads).
For this #522 PR that just means staying floor-neutral: pinned mypy, the one-line assert, typing_extensions under TYPE_CHECKING.

Sorry for being stubborn about this, tell me if I'm missing something...

pyproject.toml
mistotebe: How about this? Still depending on `typing_extensions` though, I'd argue that bullseye nor jammy will pick it up themselves and people who build from source would be able to install it just fine?

Lib/ldap/ldapobject.py
mistotebe: Ok, I've put the 3.9 floor back, can you move the rest of the comment that's relevant to #641 there?

mistotebe:
> Sorry, that was too terse..:D On main, Entry.attribute_types() worked when the objectClass values were str and blew up on bytes. Now it decodes them (models.py:795 - `ocs = [oc.decode("utf-8") for oc in bin_ocs]`), so bytes work, which is what search results gives you, but str values fail with AttributeError on .decode. IMO, decode only when it's bytes and both keep working.

The typing on the Entry `__getitem__` suggests there is no way to get anything but `bytes` on the value side but I've tweaked it to accept both just in case people manage to push things in right now.

> > > ```
> > > * the _s methods return None as documented,

How about this one?

> > > * the default syncrepl cookie is '' instead of None,
> > > ```
> > 
> > Sound like something that should not change, can you show me where?
> 
> syncrepl.py:527, syncrepl_get_cookie() returns `''` where main returned `None`. syncrepl_search() passes it straight into SyncRequestControl (:406-408) and encodeControlValue adds the syncCookie element whenever the cookie is not None (:107). So a consumer that doesn't override the hook now sends an empty cookie instead of none. I'd keep returning None there (-> str | None).

I've adjusted accordingly. Might be too loose but arguably, we do take `str` in so many places that should take `bytes` that it's here to stay, at least here.

droideck (dismissed):
LGTM! Thank a lot for finishing the change!
Great work.

PR #527, opened 2023-06-14
Add retry on EINTR.

mit-da:
Try and fix issue #242 by adding a retry on EINTR and a sleep before attempting to do so. This is based off of https://github.com/python-ldap/python-ldap/pull/291. 
We happen to come across a few instances of EINTR in our usage every day and would like to add a single retry since it seems all further LDAP calls are successful. Only kept it to a single retry for now, though I could be convinced to raise it.

### DIFF
--- a/Lib/ldap/ldapobject.py
+++ b/Lib/ldap/ldapobject.py
@@ -31,7 +31,8 @@ from ldap import LDAPError
 from errno import EINTR
 
 def _retry_on_interrupted_ldap_call(func, *args, **kwargs):
-  """Call func, retrying if it raises an LDAPError with errno == EINTR.
+  """
+  Call func, retrying if it raises an LDAPError with errno == EINTR.
   """
 
   attempts = 0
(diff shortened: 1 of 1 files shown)


mit-da:
Apologies for the delay, pretty inexperienced with python. Made some changes to fix tests.

spaceone:
doesn't it already work when you use `ldap.ldapobject.ReconnectLDAPObject` instead of `ldap.ldapobject.SimpleLDAPObject`?

mit-da:
I tested with ldap.ldapobject.ReconnectLDAPObject and unfortunately the interrupts still occur and don't seem to retry. The bind action seems to work but a call to say modify a group will throw the error.

mistotebe:
Is setting the `LDAP_OPT_RESTART` option not working? Can you post the backtrace when this happens?

mistotebe:
-1 until we know if/when `LDAP_OPT_RESTART` is (not) enough. Don't want to paper over potential libldap bugs.

mistotebe:
Assuming the absence of `LDAP_OPT_RESTART` was the culprit.

PR #537, opened 2023-10-02
Test with Python 3.12

tiran:
- Add 3.12 to setup.cfg, tox, and GHA
- Use Ubuntu 22.04 except for Python 3.6
- Add weekly cronjob to run CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,13 @@
 ---
 name: CI
 
-on: [push, pull_request]
+on: 
+  push:
+  pull_request:
+  schedule:
+    # every Monday
+    - cron: '30 4 * * 1'
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


tiran:
The `%f` format operator was never supported. It's now raising an error in 3.12

https://docs.python.org/3.12/c-api/unicode.html#c.PyUnicode_FromFormat

droideck (approved):
Looks good! Ack

tiran:
@droideck Could you please reconfigure CI to require `Ubuntu with Python pypy3.9` instead of `Ubuntu with Python pypy3`? `actions/setup-python@v4` requires a minor version for PyPy, too.

droideck:
> @droideck Could you please reconfigure CI to require `Ubuntu with Python pypy3.9` instead of `Ubuntu with Python pypy3`? `actions/setup-python@v4` requires a minor version for PyPy, too.

Do you mean to modify your PR and change https://github.com/python-ldap/python-ldap/blob/main/.github/workflows/tox-fedora.yml ?
Because for ci.yml, I already see the change in this PR.

Or is there something in GitHub project settings that only maintainers can access? If so, I wasn't able to find... Any guidance is appreciated. :)

tiran:
> Or is there something in GitHub project settings that only maintainers can access? If so, I wasn't able to find... Any guidance is appreciated. :)

It's a GitHub setting somewhere under branch protection / required status checks. `Ubuntu with Python pypy3` is a required test, but the test no longer exists.

droideck:
> > Or is there something in GitHub project settings that only maintainers can access? If so, I wasn't able to find... Any guidance is appreciated. :)
> 
> It's a GitHub setting somewhere under branch protection / required status checks. `Ubuntu with Python pypy3` is a required test, but the test no longer exists.

That was it! Hidden under the branch `edit` button... Thanks!

Issue #540, opened 2023-10-30
[feature] Port python-ldap to PEP 630 (Isolating Extension Modules)


tiran:
I propose to port python-ldap to [PEP 630](https://peps.python.org/pep-0630/). Actually, I have already done the work and have a working prototype. PEP 630 is an informational PEP that covers three PEPs

* [PEP 384](https://peps.python.org/pep-0384/) – Defining a Stable ABI
* [PEP 489](https://peps.python.org/pep-0489/) – Multi-phase extension module initialization
* [PEP 573](https://peps.python.org/pep-0573/) – Module State Access from C Extension Methods

Isolated C extension module will give us two major benefits:

* Stable ABI / limited API will allow us / our users to create `abi3` binary wheels. My prototype implements `cp36-abi3-linux_x86_64.whl` support. The wheels is compatible with all Python versions from 3.6 to 3.12, as well as future versions of Python 3 with ABI3. For example @cgohlke won't need to create Windows binaries for each Python version.
* It makes python-ldap compatible with [PEP 684 Per-Interpreter GIL](https://peps.python.org/pep-0684/) and safe to use with multiple subinterpreters.

## Required changes

- [ ] Port `PyTypeObject LDAP_Type` to heap type.
  - Replace static `PyTypeObject LDAP_Type` with `PyType_Slot` and `PyType_Spec`
  - Allocate object with `PyObject_GC_New` and free it with `PyObject_GC_UnTrack` / `PyObject_GC_Del`
  - Add GC protocol (traverse, clear) to track the heap reference to the type
  - Initialize type with `PyType_FromSpec`
- [ ] Create `_ldap` C extension module with multi-phase initialization
  - Export module-level functions from `functions.c` and `ldapcontrol.c`. This also gets rid of `LDAPadd_methods` hack, `LDAPinit_functions`, and `LDAPinit_control`
  - Define `PyModuleDef_Slot` to init type, constants, and pkginfo
  - Make `PyModuleDef` global static
  - Initialize module with `PyModuleDef_Init`
- [ ] Remove static globals `LDAP_Type`, `LDAPexception_class`, and `errobjects` in favor of module state struct. 
  - Define `LDAPModState` and add it to `PyModuleDef`
  - Add state pointer to `LDAPObject`. `PyType_GetModuleState()` is not available in Python < 3.9 and became part of the stable ABI in 3.10.
[...]

PR #541, opened 2023-10-30
refactor: Merge all header files

tiran:
Merge all header files except `constants_generated.h` into a single header file `pythonldap.h`. A single header file makes it far easier to port python-ldap to heap types and module state for Per-Interpreter GIL.

`pythonldap.h` uses new macros `PYLDAP_FUNC` and `PYLDAP_DATA` to declare functions and data, which are used across C files.

Remove unused macro `streq`.

Remove Python 2 vestiges from C and header files.


droideck (approved):
Looks good!

PR #542, opened 2023-10-30
refactor: Use functions from limited API

tiran:
Replace unsafe macros and direct struct access with functions from the subset of limited API functions.

* `PySequence_Fast_GET_ITEM` -> `PySequence_GetItem`
* `PyTuple_SET_ITEM` -> `PyTuple_SetItem`
* `PyList_SET_ITEM` -> `PyList_SetItem`
* `PyUnicode_AsUTF8AndSize` -> `PyUnicode_AsUTF8String` + `PyBytes_AsStringAndSize`. The function `PyUnicode_AsUTF8AndSize` is not in limited API before Python 3.10.
* `const char *tp_name` -> `Py_TYPE()` string representation

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -280,9 +280,9 @@ attrs_from_List(PyObject *attrlist, char ***attrsp)
     }
     else {
         PyObject *item = NULL;
+        PyObject *bstr;
         Py_ssize_t i, len, strlen;
-
-        const char *str;
+        char *str;
 
         seq = PySequence_Fast(attrlist, "expected list of strings or None");
         if (seq == NULL)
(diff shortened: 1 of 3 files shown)


mistotebe:
Shouldn't this also enable the limited API so the changes are exercised immediately?

tiran:
This PR only replaces a few functions that are outside the limited API set. I have split my original work into smaller chunks, that are easier to understand and review. There is a bunch of more work to do before I can enable the limited API, e.g. port LDAP_Type to heap type.

mistotebe:
> This PR only replaces a few functions that are outside the limited API set. I have split my original work into smaller chunks, that are easier to understand and review. There is a bunch of more work to do before I can enable the limited API, e.g. port LDAP_Type to heap type.

Right I assumed it was considered ready for review, not a draft?

tiran:
> Right I assumed it was considered ready for review, not a draft?

I don't follow. This PR is not a draft and is ready for review.

I'm sorry if I wasn't clear. I want to split the work on #540 into small, digestible chunks. Each PR will improve and change one aspect of the code base. Every PR will also leave the code base in working conditions with all tests passing. 

This approach makes it easier for you and @droideck to review and verify the changes. Some changes are not obvious unless you are familiar with stable ABI and module states. I estimate that #540 is going to take around 6 to 7 PRs in total.

mistotebe:
Hmm, sure, but at this point how do I tell whether the PR is useful? It gives no clear way to see if something has been missed for one.

I can see #540 is large and you want to split it into different parts (and get your point there) but I'd still expect each PR to be a patchset with a clear intent along the path that gets realised (somewhat) by the end of it, not necessarily one patch per PR if that makes sense? This PR as it stands looks like one commit that can't be reviewed alone, which might change if more commits were added to it (e.g. being able to limit ourselves to the stable ABI for 3.x+ and enforcing that).

You are welcome to disagree but then I'll have to recuse myself from reviewing this.

Modules/LDAPObject.c
```diff
@@ -312,36 +308,34 @@ attrs_from_List(PyObject *attrlist, char ***attrsp)
 
         for (i = 0; i < len; i++) {
             attrs[i] = NULL;
-            item = PySequence_Fast_GET_ITEM(seq, i);
+            item = PySequence_GetItem(seq, i);
```
encukou, line 311: This returns a new reference, item needs to be decref'd.

tiran: https://github.com/python-ldap/python-ldap/compare/bad8aef9dc8c74049a73784c3bb23a1c9bbd0ce1..b45e0c3826c5c41b8e1326f1a9c46ff00bbe75df

tiran:
But this PR has a clear intent; it replaces several macros, one function, and one case of direct struct access with features from the limited API. To review the PR, you look up the documentation of each replaced [function](https://docs.python.org/3/c-api/unicode.html?highlight=pyunicode_asutf8andsize#c.PyUnicode_AsUTF8AndSize) / [macro](https://docs.python.org/3/c-api/sequence.html?highlight=pysequence_fast_get_item#c.PySequence_Fast_GET_ITEM), check that the old code is not marked for `Part of the Stable ABI`, and the new code is in the stable ABI for 3.6+.

You could argue that my patch may be missing some function replacements. That's not a problem, we can address that in another commit. The important bit is that this PR does not introduce a regression.

PR #543, opened 2023-10-30
test: PyPy fixes

tiran:
On PyPy 3.9 `sock.close()` sometimes fails with `Bad file descriptor`, because the fd is already closed. Silence the exception. The problem is not a bug in python-ldap.

Really use PyPy 3.9 in `tox.ini`.

### DIFF
--- a/Tests/t_ldapobject.py
+++ b/Tests/t_ldapobject.py
@@ -657,7 +657,11 @@ class Test03_SimpleLDAPObjectWithFileno(Test00_SimpleLDAPObject):
         )
 
     def tearDown(self):
-        self._sock.close()
+        try:
+            self._sock.close()
+        except OSError:
+            # PyPy sometimes fails with "Bad file descriptor"
+            pass
         delattr(self, '_sock')
         super().tearDown()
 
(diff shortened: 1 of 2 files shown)


mistotebe:
-1 on this: we should really figure out why the socket was closed already to an authoritative reason.

Being opportunistic and then masking the error risks hiding hard to diagnose high severity bugs which we've had a hard time with using heimdal etc. in the past (a library closes its file descriptor, the fd # is reused by independent code and the library closes that one too).

tiran:
Be my guest and good luck figuring it out... The problem only occurs with PyPy 3.9 on GitHub Actions and only every few test runs. It's rather random. I have not been able to reproduce the issue locally either.

In the mean time, I can work around the issue by restarting the failed test manually.

mistotebe:
Ok, I've had a try and can reproduce this quite reliably (even with `--jit off`, I think). Indeed close() is called several times on the same fd in Test03_SimpleLDAPObjectWithFileno, however trying to debug this in gdb, I cannot see **who** is calling the close() in each instance. Unlike with CPython I can't seem to be able to extract the python stack at the time of that call. Have you had any success/do you have hints on how that could be achieved?

PR #544, opened 2023-11-01
Claim ownership of socket once we've passed it to libldap

mistotebe:
Per [Python docs](https://docs.python.org/3/library/socket.html#socket.socket.detach) a `socket` object will close the underlying file descriptor on garbage collection and to use it elsewhere, one needs to call `detach()`.

Fixes: https://github.com/python-ldap/python-ldap/issues/460
Closes: https://github.com/python-ldap/python-ldap/pull/543


Tests/t_ldapobject.py
```diff
@@ -649,22 +649,14 @@ class Test03_SimpleLDAPObjectWithFileno(Test00_SimpleLDAPObject):
     def _open_ldap_conn(self, who=None, cred=None, **kwargs):
         if hasattr(self, '_sock'):
             raise RuntimeError("socket already connected")
```
tiran, line 651: You can remove these lines, too.

tiran (approved):
Excellent work, thanks for digging into this rabbit hole.

PR #546, opened 2023-11-03
refactor: Port LDAP_Type to heap type

tiran:
The `LDAP` type has been converted from a static type to a heap type. The limited API does not support static types.

Heap types behave more closely like Python classes. They are allocated on the heap and reference counted. Instances have a strong reference to their type and must use GC protocol to track this reference.

The LDAP type can no longer be instantiated by Python code. This was never supported and resulted in an invalid LDAP connection. Code like `type(_ldap.initialize(""))()" now fails with a `TypeError`.

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -12,36 +12,75 @@
 
 static void free_attrs(char ***);
 
-/* constructor */
+/* global heap type object */
+PyTypeObject *LDAP_Type;
 
+/* constructor */
 LDAPObject *
 newLDAPObject(LDAP *l)
 {
-    LDAPObject *self = (LDAPObject *)PyObject_NEW(LDAPObject, &LDAP_Type);
+    LDAPObject *self = (LDAPObject *)PyObject_GC_New(LDAPObject, LDAP_Type);
 
-    if (self == NULL)
+    if (self == NULL) {
         return NULL;
+    }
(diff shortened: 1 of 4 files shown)


Modules/LDAPObject.c
```diff
@@ -12,36 +12,71 @@
...
 newLDAPObject(LDAP *l)
 {
-    LDAPObject *self = (LDAPObject *)PyObject_NEW(LDAPObject, &LDAP_Type);
+    LDAPObject *self = (LDAPObject *)PyObject_GC_New(LDAPObject, LDAP_Type);
 
-    if (self == NULL)
+    if (self == NULL) {
         return NULL;
+    }
+#if PY_VERSION_HEX < 0x03080000
```
encukou, line 27: This is the version of Python used at build time. If you'll use stable ABI, you want the runtime version.
Unfortunately, this gotcha was only found and fixed recently -- [in 3.11](https://docs.python.org/3/c-api/apiabiversion.html#c.Py_Version). The best workaroud I'm aware of is rather tedious -- parsing `sys.version_info` via `PySys_GetObject` and `PyArg_ParseTuple`, with a `static` (per-process) cache.

:(

Or use stable ABI only for 3.8+.

tiran: Meh, I haven't considered this gotcha. 3.6 and 3.7 have reached EOL anyway. Let's not support them for limited API.

PR #547, opened 2023-11-09
Prepare a new release

droideck:


droideck:
I plan to cherry-pick the commit to https://github.com/python-ldap/python-ldap/tree/3.4 branch, and I'll make the release from there.
The `main` branch already has additional refactoring done by @tiran - and I think the PEP 630 work should be released as a whole unit when it's ready.

@mistotebe @tiran, will you be okay with such a plan? And with this `CHANGES` for the release?

CHANGES
```diff
@@ -1,3 +1,20 @@
+Released 3.4.4 2022-11-14
+
+Fixes:
+* Reconnect race condition in ReconnectLDAPObject is now fixed
+* Socket ownership is now claimed once we've passed it to libldap
+* LDAP_set_option string formats are now compatible with Python 3.13
```
tiran, line 6: The problem affects 3.12
```suggestion
* LDAP_set_option string formats are now compatible with Python 3.12
```

tiran (dismissed):

intgr:
@cgohlke Could you please update [python-ldap-build](https://github.com/cgohlke/python-ldap-build/) to 3.4.4 too?

PR #551, opened 2023-11-20
Add readthedocs configuration file

mistotebe:
Running without one has apparently been deprecated since September 2023.

BTW: RTD builds failing is probably unrelated (there is no indication of why they're failing to say for certain).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,22 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: Doc/conf.py
+
+# We recommend specifying your dependencies to enable reproducible builds:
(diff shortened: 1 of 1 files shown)


mistotebe:
An example failure for 3.4.4 that I just can't parse: https://readthedocs.org/api/v2/build/22604489.txt

Any ideas welcome, I'll also try later just in case it's something on RTD side they'll fix.

droideck:
> An example failure for 3.4.4 that I just can't parse: https://readthedocs.org/api/v2/build/22604489.txt
> 
> Any ideas welcome, I'll also try later just in case it's something on RTD side they'll fix.

Did you have a change to test this PR on readthedocs project? (with a separate python-ldap-test project or something)? 

If I understand correctly, the https://readthedocs.org/api/v2/build/22604489.txt failure is what we have without the PR, right?

mistotebe:
No haven't been able to test this at all, adapted the sample from https://blog.readthedocs.com/migrate-configuration-v2/ - which I just noticed has a deprecation timeline ending with a complete deprecation in the past so that might explain it.

Yeah, all failures are against current tags/main.

droideck (approved):
Okay, I've tested and it looks good for the latest version (which has your commit on it)!
https://readthedocs.org/projects/python-ldap-test/builds/22637292/

Ack.

droideck:
@mistotebe I think we can merge it. Is it okay with you?

PR #566, opened 2024-04-22
Fixes #565 - Use name values instead of raw decimal

### LINKED ISSUE #565: Replace raw type values in syncrepl.py with the name values
If you found a bug in python-ldap, or would request a new feature,
this is the place to let us know.

Please describe the issue and your environment here.

---

Issue description:

The code in syncrepl.py uses raw type values (100, 101, 121) instead of their corresponding name values which makes the code less legible than necessary

Steps to reproduce: 

Inspect code

quanah:
Use the name values for result types in syncrepl.py rather than the raw decimal values.


mistotebe (approved):
LGTM

droideck (approved):
Looks good! Ack

PR #571, opened 2024-06-25
Add OpenLDAPSyncreplCookie class

mistotebe:


Lib/ldap/syncrepl.py
```diff
@@ -535,3 +536,71 @@ def syncrepl_refreshdone(self):
...
+    OpenLDAPSyncreplCookie - allows a consumer to track a cookie across a
+    refreshAndPersist syncrepl session against a multi-provider OpenLDAP cluster
+    """
+
+    rid: int = 0
+    sid: int = 0
+    _csnset: Dict[int, str]
+
+    def __init__(self, cookie: AnyStr = "") -> None:
+        self._csnset = {}
```
droideck, line 552: It's really minor, but I think it's inconsistent.
You set `self._csnset` default in `__init__` but for `self.rid` and `self.sid` - you don't. Something like this should be nicer, IMO

```
    def __init__(self, cookie: AnyStr = "") -> None:
        self.rid = 0
        self.sid = 0
        self._csnset = {}

File: Lib/ldap/syncrepl.py [hunk 2]
```diff
@@ -535,3 +536,71 @@ def syncrepl_refreshdone(self):
...
+
+        if cookie:
+            self.update(cookie)
+
+    def _parse_csn(self, csn: str) -> List[str]:
```
droideck, line 557: I think the return value can be a tuple here and we can specify the number of elements right away: ` -> Tuple[str, str, str, str]`

Also, it'll be nice to validate the tuple's len after the split.

Lib/ldap/syncrepl.py
```diff
@@ -535,3 +536,71 @@ def syncrepl_refreshdone(self):
...
+                # Did not recognise this cookie part
+                pass
+        return result
+
+    def update(self, cookie: str):
+        """
+        Update the CSN set based on a cookie we just received, use in
+        syncrepl_set_cookie() to track the session state.
+        """
+        components = self._parse_cookie(cookie)
```
droideck, line 585: What about `rid` and `sid` that we parsed (if any)? Should we also assign it here or do I miss something?

droideck (dismissed):
A few minor comments, but the code looks good!

We need to figure out, though, what's wrong with the CI... I can't even restart it.

Lib/ldap/syncrepl.py
mistotebe: Good point, I'll adjust.

Lib/ldap/syncrepl.py
mistotebe: With OpenLDAP's MPR, signalling a non-zero `sid` carries meaning and puts more responsibility on the consumer. There is no need to track the rid either with the code in python-ldap as-is, so it's zeroed out as well - doing so makes it possible to normalise the cookies if comparison (as a vector clock) is desired.

Lib/ldap/syncrepl.py
droideck: Ir was really a minor concern, and as you explained that there is no need to track the `rid` and `sid,` I'm totally okay with the way it is now.

droideck (dismissed):
LGTM!
Could you please try to rebase and see if tests pass?

mistotebe:
> LGTM! Could you please try to rebase and see if tests pass?

Just rebasing didn't seem to work, but given that @spaceone had some comments and I forgot to register the tests, this rebase seems to have passed made pypy310 happy. We might still be closing a fd twice somewhere else in the code?

Tests/t_ldap_syncrepl.py
```diff
@@ -442,6 +447,174 @@ def setUp(self):
...
+            self.server.ldap_uri,
+            self.server.root_dn,
+            self.server.root_pw,
+            bytes_mode=False
+        )
+        self.suffix = self.server.suffix
+
+        # An active MPR should not have a sid=000 server in it
+        if self.server.server_id == 0:
+            self.skip("Server got serverid 0 assigned")
```
droideck: Looking at these methods - https://docs.python.org/3/library/unittest.html#unittest.TestCase , I think it should be:

```python
            self.skipTest("Server got serverid 0 assigned")
```

the only place I see `.skip()` - is as a decorator.

Lib/ldap/syncrepl.py
```diff
@@ -535,3 +536,71 @@ def syncrepl_refreshdone(self):
...
+            elif part.startswith('csn='):
+                result['csn'] = part[4:].split(';')
+            elif part.startswith('delcsn='):
+                result['delcsn'] = part[7:]
+            else:
+                # Did not recognize this cookie part
+                pass
+        return result
+
+    def update(self, cookie: str):
```
droideck: The signature says cookie: str but _parse_cookie() accepts AnyStr (both str and bytes). Worth make it consistent.

Lib/ldap/syncrepl.py
```diff
@@ -535,3 +536,71 @@ def syncrepl_refreshdone(self):
         follows.
         """
         pass
+
+
+class OpenLDAPSyncreplCookie:
```
droideck, line 542: Should we also export the class in `__all__`?

droideck (dismissed):
Very minor comments, otherwise, LGTM!

mistotebe:
Thanks @droideck for spotting those, all of them should have been there in the first place. Sorry about that, should be fixed now.

droideck (approved):
Sure! And LGTM:)

PR #583, opened 2025-01-29
Deprecate EOL Python 3.6, 3.7, 3.8 versions

droideck:


michaelalang:
oh damn ... I just added more logic to honor the older versions :D ... all good just joking <3

mistotebe:
Wasn't @tiran suggesting we had to keep 3.6 around because RedHat was interested in supporting it for RHEL8 sake? Does it mean their position has changed now?

droideck:
> Wasn't @tiran suggesting we had to keep 3.6 around because RedHat was interested in supporting it for RHEL8 sake? Does it mean their position has changed now?

For RHEL 8 (or any other older distribution), some time ago, I created [3.4 branch](https://github.com/python-ldap/python-ldap/tree/3.4). So if any security or other issues arise, we can cherry-pick the fixes to 3.4, and we can make the 3.4.5 release there.

In the meantime, soon, I think we can do the 4.0 release and include this commit (and any other more significant changes we might want, like syncrepl support - but we can discuss it separately)

What do you think?

michaelalang:
@mistotebe cool and no worries ... anything left from myside to do ? since I am not a project contributor I assume it's not possible to merge it myself :D

PR #589, opened 2025-07-13
Package python-ldap with pyproject.toml

droideck:


droideck:
@hroncok hi! When/if you have time, could you please take a look if all is good with the `pyproject.toml` transformation?

Notably, it has C extensions that are handled in `setup.py` for now.
I've tried to make it work with `pyproject.toml` but I faced this issue - https://github.com/pypa/setuptools/issues/4810

INSTALL
```diff
@@ -1,8 +1,8 @@
 Quick build instructions:
 
     edit setup.cfg (see Build/ for platform-specific examples)
-    python setup.py build
-    python setup.py install
+    python -m build
+    pip install .
```
hroncok, line 5: You shouldn't need both commands.

pyproject.toml
```diff
@@ -1,4 +1,151 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
```
hroncok, line 4:
```suggestion
```

- current version of setuptools (70.1+) does not need wheel at all
- older versions of setuptools would fetch wheel when building wheels (but not sdists)

droideck:
> I would also suggest not mixing unrelated changes, but converting this to pyproject.toml as is and only then add more stuff (like ruff).

I totally agree! It's better to be done in a separate PR.

mistotebe:
I support the changes in general. As @hroncok pointed out, it's worth doing it one commit at a time (e.g. cleanup if needed, then switch to pyproject.toml as 1:1, then any changes to 3.13 if we really need to do them **right now**)

The doc updates look quite useful, so personally I'd accept them as part of the cleanup set but others' opinions might differ. Adding `ruff` or other modernisation (`noxfile.py`?) sounds like a new PR though.

PR #596, opened 2025-10-07
Prepare a new release

droideck:
Security fixes:
* CVE-2025-61911 (GHSA-r7r6-cc7p-4v5m): Enforce ``str`` input in
  ``ldap.filter.escape_filter_chars`` with ``escape_mode=1``; ensure proper
  escaping. (thanks to lukas-eu)
* CVE-2025-61912 (GHSA-p34h-wq7j-h5v6): Correct NUL escaping in
  ``ldap.dn.escape_dn_chars`` to ``\00`` per RFC 4514. (thanks to aradona91)

Fixes:
* ReconnectLDAPObject now properly reconnects on UNAVAILABLE, CONNECT_ERROR
  and TIMEOUT exceptions (previously only SERVER_DOWN), fixing reconnection
  issues especially during server restarts
* Fixed syncrepl.py to use named constants instead of raw decimal values
  for result types
* Fixed error handling in SearchNoOpMixIn to prevent a undefined variable error

Tests:
* Added comprehensive reconnection test cases including concurrent operation
  handling and server restart scenarios

Doc/
* Updated installation docs and fixed various documentation typos
* Added ReadTheDocs configuration file

Infrastructure:
* Add testing and document support for Python 3.13


droideck:
@spaceone hi!
Regarding the DN format enhancement patches https://github.com/python-ldap/python-ldap/pull/466, I've decided to postpone these until the next major release as they introduce new features and API extensions, while I'd like to keep 3.4.5 focused on bug fixes and stability improvements following semantic versioning best practices.

The most anticipated fix https://github.com/python-ldap/python-ldap/pull/267 is included, though!

Firstyear (approved):
Looks good! Thank you!

spaceone:
I understand your semantic versioning thought.
On one hand #466 is a feature but on the other hand it's a fix:

It's nearly 4 times faster, which is a performance fix.
And it allows DN normalization as the LDAP server does it.
Currently for example, we have a bug due to it, because the LDAP DNs normalized by the current Python functions which don't match the real DNs from the OpenLDAP backend, so that our Javascript frontend doesn't detect the currently set value (`cn=+1` != `cn=\2B1`) and we have no way to compare a DN correctly in Javascript.

Lib/ldap/ldapobject.py
```diff
@@ -833,6 +832,10 @@ class ReconnectLDAPObject(SimpleLDAPObject):
   * retry_delay: specifies the time in seconds between reconnect attempts.
 
   This class also implements the pickle protocol.
+
+  .. versionadded:: 3.5
```
mistotebe: 3.4.5

pyproject.toml
```diff
@@ -1,8 +1,107 @@
...
+    "Programming Language :: C",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
mistotebe, line 36: Do we support 3.13/3.14?

droideck: I've added 3.13 support as we have it on `main`. For 3.14, we can deal with it later, I think.

droideck:
> I understand your semantic versioning thought. On one hand #466 is a feature but on the other hand it's a fix:
> 
> It's nearly 4 times faster, which is a performance fix. And it allows DN normalization as the LDAP server does it. Currently for example, we have a bug due to it, because the LDAP DNs normalized by the current Python functions which don't match the real DNs from the OpenLDAP backend, so that our Javascript frontend doesn't detect the currently set value (`cn=+1` != `cn=\2B1`) and we have no way to compare a DN correctly in Javascript.

Yes... I understand that the DN normalization issue is causing real problems in your production environment. However, adding 217 lines of new C code with API extensions in a patch release goes against semantic versioning principles and risks introducing instability for other users expecting only bug fixes in 3.4.5.

We also have a couple of CVEs to include as soon as possible, so I'm a bit in a rush (even though they are Low severity) and I really want to focus this build on being a patch release.

Besides, I think 4.0 should really be released soon. 
Hopefully, we'll wrap up https://github.com/python-ldap/python-ldap/pull/464 quite soon, and I'll be happy to make 4.0 build right after that.
If @mistotebe thinks that https://github.com/python-ldap/python-ldap/pull/464 might take more time, we can release 3.5.0 sooner.
WDYT?

PR #600, opened 2025-10-21
fix(extop.dds): make passing of requestName optional again

spaceone:
requestName is already set at class member. It seems there is code out there which is not giving it as value but passing None. in that case, fallback to the class member.

Fixes: 414ae1de91543a1c0fee0738f97fe1a33d0fe666


droideck (approved):
Looks good!

One note though, if someone explicitly passes `requestName=""`, it will fallback to the class default. This is likely desired behavior, but worth noting.

We might make it more explicit:
```python
if requestName is None: requestName = self.requestName 
```

But ` requestTtl` already follows `or` patterns and I don't think any body would use it with `""`.

Issue #575, opened 2024-08-30
ldap.initialize(fileno=fileno) does take ownership of the fileno


ClydeByrdIII:
Issue description:

The [ldap.initialize() documents](https://github.com/python-ldap/python-ldap/blob/a58282adbc6b1f5f9755458227e6bb8667b72f6b/Doc/reference/ldap.rst?plain=1#L43) say
> If fileno parameter is given then the file descriptor will be used to connect to an LDAP server. The fileno must either be a socket file descriptor as [int](https://docs.python.org/3/library/functions.html#int) or a file-like object with a fileno() method that returns a socket file descriptor. The socket file descriptor must already be connected. [LDAPObject](https://www.python-ldap.org/en/python-ldap-3.4.3/reference/ldap.html#ldap.ldap.ldapobject.LDAPObject) **does not take ownership of the file descriptor**. It must be kept open during operations and explicitly closed after the [LDAPObject](https://www.python-ldap.org/en/python-ldap-3.4.3/reference/ldap.html#ldap.ldap.ldapobject.LDAPObject) is unbound.

However it seemingly is the case that it *does* take ownership

https://github.com/python-ldap/python-ldap/commit/1490e999e10960f0fdf974f6da804a2fc55e4b05, https://github.com/python-ldap/python-ldap/issues/460, https://github.com/python-ldap/python-ldap/pull/543 are likely due to this confusion as well.

Following the code shows it goes like this.

`ldap.initalize(*args, fileno=fd)` will result in `ldap_init_fd()` being called which executes [code](https://github.com/openldap/openldap/blob/c9ab732ec18e32c58643e44a45233a6ed87830b6/libraries/libldap/open.c#L329) that essentially creates a new ldap connection object that is then bound to the passed in fileno it then based on the protocol passed (LDAP_PROTO_TCP, LDAP_PROTO_UDP, LDAP_PROTO_IPC, LDAP_PROTO_EXT) will assign the protocols respective sockbuf_IO(s) set of {setup, ctrl, read, write, close} functions (except for LDAP_PROTO_EXT which the user specifies a custom struct elsewhere). These functions dictate how to interact with the underlying communication mechanism. Most relevant to this issue is the “close” functions

For TCP/UDP sockbuf_IO it calls “tcp_close()” which is system dependent macro but for Linux is likely  the [below](https://github.com/openldap/openldap/blob/c9ab732ec18e32c58643e44a45233a6ed87830b6/include/ac/socket.h#L142)
[...]

PR #603, opened 2025-11-12
CI workflow improvements

merlinz01:
- Merge both workflows into a single CI workflow
- Remove scheduled workflow run - I don't know of any benefit
- Upgrade and pin action versions to latest
- Use `uv` to speed up installing Python and dependencies
- Format and reorganize YAML structure for consistency


mistotebe:
Can you rebase on top of main? Also I'd keep at least the scheduled workflows, they're the ones that give us a better chance of exposing things that don't trigger on every run like resource mismanagement (#575).

PR #604, opened 2026-01-13
Allow for Modlist with ordered attribute values

tacerus:
In some LDAP implementations attribute ordering is relevant. Allow for modifications which consider the order of attributes by implementing an optional "strict_order_attr_types" parameter which takes a list of attributes to exempt from the transformation to a set.


tacerus:
FYI I tried to run the automatic formatting target but it reformatted lots of existing files in the repository, so I did not include that here.

PR #613, opened 2026-05-19
Miscellaneous

spaceone:


pyproject.toml
```diff
@@ -13,7 +13,7 @@ authors = [
     {name = "python-ldap project", email = "python-ldap@python.org"},
 ]
 readme = "README"
-requires-python = ">=3.6"
+requires-python = ">=3.8"
```
fumiyas: Should be 3.9?

pyproject.toml
```diff
@@ -26,8 +26,6 @@ classifiers = [
     "Programming Language :: C",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.6",
-    "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
```
fumiyas: Should be removed?

pyproject.toml
spaceone: yes, done :-)

mistotebe (approved):
LGTM

BTW, I will probably be bumping the minimum requirements a little further (probably 3.10 but could be persuaded to go up to 3.11) soon with #540

PR #618, opened 2026-07-22
fix(vlv): include contextID in subsequent VLV requests

spaceone:
We wondered about the performance and usage of the whole memory when using the VLV control in a request: 

A environment with 200.000 users and 8 GB RAM:
* normal LDAP search: 6.125s
* `SSS` + `VLV` with `PAGE_SIZE=50`: aborted after 3-4 minutes, RAM completely full
* `SSS` + `VLV` with `PAGE_SIZE=2000`: 1m36.634s (slapd Mem 65.7 %)

* SimplePagedResultsControl `PAGE_SIZE=50`:   10.394s
* SimplePagedResultsControl `PAGE_SIZE=2000`: 6.248s

The VLV request control did not encode the optional `contextID` field, even when it was present on the request control object.

The LDAP server uses the `contextID` returned in a VLV response to continue an existing virtual list search. Omitting it causes each subsequent request to be treated as a new VLV search, preventing state reuse and leading to significantly increased processing time and memory usage on the LDAP server.

Encode `contextID` whenever it is available to allow servers to resume an existing VLV search context as intended by RFC 3876.

### DIFF
--- a/Lib/ldap/controls/vlv.py
+++ b/Lib/ldap/controls/vlv.py
@@ -86,6 +86,8 @@ class VLVRequestControl(RequestControl):
         else:
             raise NotImplementedError
         p.setComponentByName('target', target)
+        if self.context_id is not None:
+            p.setComponentByName('contextID', self.context_id)
         return encoder.encode(p)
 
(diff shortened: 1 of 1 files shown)


mistotebe (approved):
LGTM

PR #625, opened 2026-08-26
Fixing reference leaks

mistotebe:
(Might we willing to squash the series into a single commit.)

Picking up items from issues 617-622, plus independently anything picked up by the following coccinelle patch (minus false positives):

```cocci
/*
 * Run as
 * spatch --dir Modules -D report --no-show-diff --very-quiet --sp-file leaks.cocci
 */

virtual report

@r exists@
identifier alloc;
PyObject *obj;
identifier label;
position p;
@@

obj = alloc(...)
... when != return ...;
    when != obj;
(
  if (<+... obj == NULL ...+>) { ... return ...; }
|
  if (<+... obj == NULL ...+>) return ...;
|
  if (<+... obj == NULL ...+>) { ... goto label; }
|
  if (<+... obj == NULL ...+>) goto label;
)
... when != Py_DECREF(obj)
    when != Py_XDECREF(obj)
    when != Py_CLEAR(obj)
    when != obj = ...
    when != return obj;
(
  return obj;
|
  return@p ...;
)

@script:python depends on report@
p << r.p;
obj << r.obj;
alloc << r.alloc;
@@

msg = "leaked '%s' from %s()" % (obj, alloc)
coccilib.report.print_report(p[0], msg)

```

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -170,6 +170,7 @@ Tuple_to_LDAPMod(PyObject *tup, int no_op)
                 LDAPerror_TypeError
                     ("Tuple_to_LDAPMod(): expected a byte string in the list",
                      item);
+                Py_DECREF(item);
                 goto error;
             }
             lm->mod_bvalues[i]->bv_len = PyBytes_Size(item);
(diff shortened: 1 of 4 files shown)


spaceone:
I like if the commits are explicit and target one issue instead of squashing everything together.

PR #628, opened 2026-08-27
fix(ldapurl): escape HTML chars in created HTML anchor

spaceone:
to guard against Cross Site Scripting (XSS) attacks.


mistotebe (approved):
We had that in the ldapurl module? Thank you!

PR #634, opened 2026-09-04
fix(_ldap): plug error handling leaks

mistotebe:
Adjustments to #466, `_free_dn_structure()` is tolerant to partially populated structures, so optimistic registration is viable and simplifies error handling.

### DIFF
--- a/Modules/functions.c
+++ b/Modules/functions.c
@@ -251,6 +251,7 @@ l_ldap_dn2str(PyObject *self, PyObject *args)
             PyErr_NoMemory();
             goto error_cleanup;
         }
+        dn[i] = rdn;
 
         for (j = 0; j < navas; j++) {
             py_ava_item = PySequence_GetItem(py_rdn_seq, j);  /* New reference */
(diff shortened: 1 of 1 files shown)


Modules/functions.c
```diff
@@ -268,6 +269,7 @@ l_ldap_dn2str(PyObject *self, PyObject *args)
             py_encoding = PyTuple_GetItem(py_ava_item, 2);  /* Borrowed reference */
 
             if (!PyUnicode_Check(py_name) || !PyUnicode_Check(py_value) || !PyLong_Check(py_encoding)) {
+                dn[i] = rdn;
```
droideck: I'm not sure we need it as there is one already at 254 line.
Possibly, a leftover from an earlier draft?

droideck (dismissed):
Besides a small nit, looks good!

Modules/functions.c
mistotebe: You're right, thanks!

droideck (approved):

PR #630, opened 2026-09-02
fix(_ldap): do not ignore ldap_parse_result's errors

mistotebe:

PR #635, opened 2026-09-04
fix(_ldap): Do not leak retoid

### LINKED ISSUE #632: ldap_parse_extended_result(..., &retoid, ...) leaks
Some extended results/intermediate responses can omit an OID (RFC4511 allows it). l_ldap_result4 doesn't know how to handle this situation.

mistotebe:
Fixes: https://github.com/python-ldap/python-ldap/issues/632

### DIFF
--- a/Modules/LDAPObject.c
+++ b/Modules/LDAPObject.c
@@ -1058,8 +1058,8 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
     LDAPMessage *msg = NULL;
     PyObject *retval, *pmsg, *pyctrls = 0;
     int res_msgid = 0;
-    char *retoid = 0;
-    PyObject *valuestr = NULL;
+    struct berval *retdata = NULL;
+    char *retoid = NULL;
     int result = LDAP_SUCCESS;
     LDAPControl **serverctrls = 0;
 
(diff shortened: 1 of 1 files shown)


Modules/LDAPObject.c
```diff
@@ -1117,17 +1117,10 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
...
 
         if (res_type == LDAP_RES_EXTENDED) {
-            struct berval *retdata = 0;
-
             LDAP_BEGIN_ALLOW_THREADS(self);
             rc = ldap_parse_extended_result(self->ldap, msg, &retoid, &retdata,
                                             0);
             LDAP_END_ALLOW_THREADS(self);
-            /* handle error rc!=0 here? */
-            if (rc == LDAP_SUCCESS) {
```
droideck, line 1127: IIUC, `rc` is assigned twice but never read now. Which leads to `-Wunused-but-set-variable`.

Modules/LDAPObject.c
```diff
@@ -1138,7 +1131,8 @@ l_ldap_result4(LDAPObject *self, PyObject *args)
 
     if (result != LDAP_SUCCESS) {       /* result error */
         ldap_controls_free(serverctrls);
-        Py_XDECREF(valuestr);
+        ber_memfree(retoid);
```
droideck, line 1134: Probably, it's better to use `ldap_memfree` for consistency and as per OpenLDAP docs for OID string.

droideck (approved):
Couple of minor issues, otherwise LGTM!

Modules/LDAPObject.c
mistotebe: Yes, that's what `man ldap_parse_extended_result` documents, so used to living inside libldap that uses the ber_ variants I guess.

Modules/LDAPObject.c
mistotebe: You're right, I've now put this into #630 and will close this PR.

mistotebe:
Closing in favour of #630

PR #639, opened 2026-09-07
test(ldap.controls): test control options against what get_option promises

### LINKED ISSUE #102: SearchNoOpControlValue: SubstrateUnderrunError: Short octet stream on tag decoding
While I was working on PR #101 I noticed an issue with decoding of control structures. The tests are now marked as expected failures. We need to investigate later:

```
======================================================================
ERROR: test_client_controls (__main__.TestLDAPObjectOptions)
----------------------------------------------------------------------
  File "Tests/t_ldap_options.py", line 136, in test_client_controls
    self._test_controls(ldap.OPT_CLIENT_CONTROLS)
  File "Tests/t_ldap_options.py", line 75, in _test_controls
    self._check_option(option, TEST_CTRL, TEST_CTRL_EXPECTED)
  File "Tests/t_ldap_options.py", line 41, in _check_option
    new = self.get_option(option)
  File "Tests/t_ldap_options.py", line 128, in get_option
    return self.conn.get_option(option)
  File "/home/heimes/dev/redhat/python-ldap/Lib/ldap/ldapobject.py", line 855, in get_option
    result = DecodeControlTuples(result)
  File "/home/heimes/dev/redhat/python-ldap/Lib/ldap/controls/__init__.py", line 150, in DecodeControlTuples
    control.decodeControlValue(encodedControlValue)
  File "/home/heimes/dev/redhat/python-ldap/Lib/ldap/controls/openldap.py", line 37, in decodeControlValue
    decodedValue,_ = decoder.decode(encodedControlValue,asn1Spec=self.SearchNoOpControlValue())
  File "/usr/lib/python3.6/site-packages/pyasn1/codec/ber/decoder.py", line 963, in __call__
    'Short octet stream on tag decoding'
pyasn1.error.SubstrateUnderrunError: Short octet stream on tag decoding
```

droideck:
OPT_SERVER_CONTROLS and OPT_CLIENT_CONTROLS hold request controls, but LDAPObject.get_option() decodes them with the response classes. Only controls with identical request and response schemas round-trip; SearchNoOpControl cannot, so drop the expectedFailure markers and test the actual behaviour instead.

Fixes: https://github.com/python-ldap/python-ldap/issues/102

### DIFF
--- a/Tests/t_ldap_options.py
+++ b/Tests/t_ldap_options.py
@@ -7,10 +7,12 @@ os.environ['LDAPNOINIT'] = '1'
 from slapdtest import SlapdTestCase, requires_tls
 
 import ldap
-from ldap.controls import RequestControlTuples
+from ldap.controls import RequestControl, RequestControlTuples
 from ldap.controls.pagedresults import SimplePagedResultsControl
 from ldap.controls.openldap import SearchNoOpControl
 from ldap.ldapobject import SimpleLDAPObject
+
+from pyasn1.error import PyAsn1Error
 
 
 SENTINEL = object()
@@ -181,15 +183,46 @@ class TestLDAPObjectOptions(BaseTestOptions, SlapdTestCase):
         finally:
             self.set_option(option, old)
 
-    # test is failing with:
-    # pyasn1.error.SubstrateUnderrunError: Short octet stream on tag decoding
-    @unittest.expectedFailure
-    def test_client_controls(self):
-        self._test_controls(ldap.OPT_CLIENT_CONTROLS)
+    # These options hold request controls, but get_option() decodes them
+    # with the response classes in KNOWN_RESPONSE_CONTROLS, so only
+    # controls with identical request and response schemas round-trip.
+    # Returning raw tuples like the module-level API is a 4.0 question, #643.
+    def _test_controls(self, option):
+        self._check_option(option, [])
 
-    @unittest.expectedFailure
-    def test_server_controls(self):
-        self._test_controls(ldap.OPT_SERVER_CONTROLS)
+        self.set_option(option, [
+            SimplePagedResultsControl(criticality=0, size=5, cookie=b'cookie'),
+        ])
+        try:
+            paged, = self.get_option(option)
+            self.assertIsInstance(paged, SimplePagedResultsControl)
+            self.assertEqual(paged.criticality, 0)
+            self.assertEqual(paged.size, 5)
+            self.assertEqual(paged.cookie, b'cookie')
+        finally:
+            self.set_option(option, [])
+
+        # no request value, SEQUENCE response: critical raises,
+        # non-critical is dropped
+        self.set_option(option, [SearchNoOpControl(criticality=1)])
+        try:
+            with self.assertRaises(PyAsn1Error):
+                self.get_option(option)
+        finally:
+            self.set_option(option, [])
+        self.set_option(option, [SearchNoOpControl(criticality=0)])
+        try:
+            self.assertEqual(self.get_option(option), [])
+        finally:
+            self.set_option(option, [])
+
+        with self.assertRaises(TypeError):
+            self.set_option(option, object)
+        with self.assertRaises(TypeError):
+            # data must be bytes or None
+            self.set_option(option, [
+                RequestControl(TEST_CTRL[0][0], TEST_CTRL[0][1], 'data'),
+            ])
 
 
 if __name__ == '__main__':


mistotebe:
You are confusing a request control 1.3.6.1.4.1.4203.666.5.18 and response control 1.3.6.1.4.1.4203.666.5.18. They have a different schema, our API is a client API, so we expect to encode request controls and decode response controls only. What the tests do is wrong, they're enconing a request control and trying to decode it as a response control, that's always going to fail.

droideck:
> You are confusing a request control 1.3.6.1.4.1.4203.666.5.18 and response control 1.3.6.1.4.1.4203.666.5.18. They have a different schema, our API is a client API, so we expect to encode request controls and decode response controls only. What the tests do is wrong, they're enconing a request control and trying to decode it as a response control, that's always going to fail.

Yeah, agreed. What I chased was the SubstrateUnderrunError out of decodeControlValue() that the expectedFailure comment points at. That was short sighted...
Reworked into a test-only change.

FWIW this isn't new. I found an interesting 2011 question about PasswordPolicyControl, same get_option -> DecodeControlTuples -> decodeControlValue path: https://mail.python.org/pipermail/python-ldap/2011q4/003037.html
With Michael answering. 

About the issue, IIUC, libldap keeps these options as request controls and hands back a duplicate (options.c does ldap_controls_dup(lo->ldo_sctrls)), and that slot is only read by ldap_int_put_controls() when encoding an outgoing request, so a response control never lands there. Module-level ldap.get_option() already returns the raw tuples.

Should the object-level one just do the same, so nothing ever decodes a request control? Is it something relevant for the https://github.com/python-ldap/python-ldap/pull/464 work you do?

mistotebe:
Personally I have never found much use for the ldap_set_option(SERVER_CONTROLS) API which attaches it to every request. The only use so far has been SessionTrackingControl which is probably the only one you do want to keep around forever. Or possibly ProxyAuthzControl but that needs a lot of care so never bothered.

mistotebe:
> FWIW this isn't new. I found an interesting 2011 question about PasswordPolicyControl, same get_option -> DecodeControlTuples -> decodeControlValue path: https://mail.python.org/pipermail/python-ldap/2011q4/003037.html
With Michael answering.

Yes, ldap_set_option(SERVER_CONTROLS) was confusing mostly for the reason that there used to be no way to extract a control from a failed operation where we would raise a naked LDAPError with that the only thing left to explore. Any leftover confusion should probably be sorted at the documentation side of things.

> Should the object-level one just do the same, so nothing ever decodes a request control?

Makes sense.

> Is it something relevant for the https://github.com/python-ldap/python-ldap/pull/464 work you do?

Yes and no, now that you get message/response objects from the result() endpoint, you don't have to fish for the controls anymore and they are right where you'd want them (entry, result, ...) all the time. Raising is the job of `requests`-like `result.raise_for_result()` getting rid of what I feel was a design mistake.

Tests/t_ldap_options.py
```diff
@@ -181,15 +183,46 @@ def test_network_timeout_attribute(self):
...
-    # test is failing with:
-    # pyasn1.error.SubstrateUnderrunError: Short octet stream on tag decoding
-    @unittest.expectedFailure
-    def test_client_controls(self):
-        self._test_controls(ldap.OPT_CLIENT_CONTROLS)
+    # These options hold request controls, but get_option() decodes them
+    # with the response classes in KNOWN_RESPONSE_CONTROLS, so only
+    # controls with identical request and response schemas round-trip.
+    # Returning raw tuples like the module-level API is a 4.0 question, #XXX.
+    def _test_controls(self, option):
```
mistotebe, line 190: Does anything call this now?

droideck: Yep, the two test methods are inherited: BaseTestOptions.test_client_controls / test_server_controls call self._test_controls(), and TestLDAPObjectOptions now overrides just the helper instead of the two methods. Both run and pass for the LDAPObject case now, where they used to be the expected failures.

```
  test_client_controls (t_ldap_options.TestGlobalOptions.test_client_controls) ... ok
  test_server_controls (t_ldap_options.TestGlobalOptions.test_server_controls) ... ok
  test_client_controls (t_ldap_options.TestLDAPObjectOptions.test_client_controls) ... ok
  test_server_controls (t_ldap_options.TestLDAPObjectOptions.test_server_controls) ... ok
  ```

File: Doc/reference/ldap.rst [hunk 2]
```diff
@@ -220,6 +230,13 @@ the following option identifiers are defined as constants:
 
 .. py:data:: OPT_SERVER_CONTROLS
 
+   Server-side request controls that libldap attaches to every operation
+   on the connection, for example
```
mistotebe, line 234:
```suggestion
   Server-side request controls that libldap attaches to every operation
   on the connection unless instructed otherwise (e.g. with
   `search_ext(..., serverctrls=[...], ...)`), for example
```

mistotebe (approved):

mistotebe:
LGTM, did you want this for 3.4.8?

droideck:
> LGTM, did you want this for 3.4.8?

Yeah, with the doc changes, I think it makes sense.
Thank you!

PR #641, opened 2026-09-08
Port module to Stable ABI + multi-phase module initialization

### LINKED ISSUE #540: [feature] Port python-ldap to PEP 630 (Isolating Extension Modules)

mistotebe:
Fixes: https://github.com/python-ldap/python-ldap/issues/540


Modules/LDAPObject.c
```diff
@@ -301,15 +337,20 @@ attrs_from_List(PyObject *attrlist, char ***attrsp)
...
             if (item == NULL)
                 goto error;
             if (!PyUnicode_Check(item)) {
                 LDAPerror_TypeError
                     ("attrs_from_List(): expected string in list", item);
+                Py_DECREF(item);
                 goto error;
             }
             str = PyUnicode_AsUTF8AndSize(item, &strlen);
+            Py_DECREF(item);
```
droideck, line 350: I think we should move this decref after `memcpy `and release `item` on the error paths too.
`str` points into `item`, so IIUC releasing our reference before copying leaves the buffer dependent on the caller’s list remaining unchanged during free-threaded execution.

Modules/ldapmodule.c
```diff
@@ -5,66 +5,88 @@
...
+    {Py_mod_exec, LDAPMod_init_constants},
+    {Py_mod_exec, init_pkginfo},
+    {0, NULL}
 };
 
-/* module initialisation */
+static struct PyModuleDef ldap_moduledef = {
+    .m_base = PyModuleDef_HEAD_INIT,
+    .m_name = "_ldap",
+    .m_size = sizeof(LDAPModState),
```
droideck, line 67: IIUC, this module/type reference cycle needs m_traverse, m_clear, and m_free so GC can collect it.
And add_err from constants.c similarly might need to release the previous array entry when another exception shares its error code.

droideck (changes requested):
I have a couple of concerns but otherwise, looks great!

Modules/constants.c
```diff
@@ -210,58 +234,53 @@ LDAPinit_constants(PyObject *m)
...
+        if (PyObject_SetAttrString(exc, "errnum", nobj) != 0) { \
+            Py_DECREF(nobj); \
+            Py_DECREF(exc); \
+            goto error; \
+        } \
         Py_DECREF(nobj); \
-        Py_DECREF(exc); \
-        goto error; \
+        state->errobjects[LDAP_##n+LDAP_ERROR_OFFSET] = exc; \
+        Py_INCREF(exc); \
```
droideck: I think the alias path now skips Py_INCREF(exc) while PyModule_AddObject() still consumes a reference...
IIUC, we need to move the incref outside the `if` so both paths provide an owned reference and cleanup doesn’t over-release the class. Or do I miss something?

droideck (approved):
Looks great!
Really happy to see the change:)

PR #642, opened 2026-09-08
Prepare a new release

mistotebe:


CHANGES
```diff
@@ -1,5 +1,20 @@
 Released 3.4.8 (unreleased)
```
droideck, line 1: The date is missing (but maybe you wanted to handle that later at the merge time)

droideck (changes requested):
Looks good!
A couple of things before the merge and a tag:

- 3.4.6 and 3.4.7 release commits also bumped the version numbers in `__version__` files  accordingly. So I think we need to do it here too.
- I also noticed we don't have SECURITY.md, CONTRIBUTING.rst, README.rst, .readthedocs.yaml, Doc/requirements.txt and Misc/python-ldap.supp (the Makefile's valgrind target uses the last one) in **MANIFEST.in**. And they were released in https://pypi.org/project/python-ldap/3.4.7/#files so I suspect the content just was in your venv when you were doing the build. IMO, worth adding the files explicitly so we can release with `python -m build` in isolation.

CHANGES
mistotebe: Exactly

mistotebe:
> Looks good! A couple of things before the merge and a tag:
> 
>     * 3.4.6 and 3.4.7 release commits also bumped the version numbers in `__version__` files  accordingly. So I think we need to do it here too.

True, I forgot we did that in the PR, and it makes more sense so done.

>     * I also noticed we don't have SECURITY.md, CONTRIBUTING.rst, README.rst, .readthedocs.yaml, Doc/requirements.txt and Misc/python-ldap.supp (the Makefile's valgrind target uses the last one) in **MANIFEST.in**. And they were released in https://pypi.org/project/python-ldap/3.4.7/#files so I suspect the content just was in your venv when you were doing the build. IMO, worth adding the files explicitly so we can release with `python -m build` in isolation.

I didn't realise they would actually land in the source package. OK, done.

droideck (approved):
LGTM!

PR #646, opened 2026-09-11
feat(syncrepl): add reloadHint parameter to syncrepl_search

Ripolin:
Expose the Sync Request Control's reloadHint flag (RFC 4533, section 2.2) through SyncreplConsumer.syncrepl_search(). When set, the server is asked to perform a full reload of the content instead of failing with e-syncRefreshRequired when it cannot resume the session from the provided cookie.

The parameter defaults to False, preserving the current behaviour.


droideck (approved):
LGTM!

Issue #647, opened 2026-09-11
Clear up str|bytes confusion where bytes are needed


mistotebe:
A lot of interfaces are confused what needs to be returned. Sometimes we decode what should stay bytes (e.g. pre/post read controls, syncrepl cookies), sometimes we return bytes where input should/must be a str (DNs, ...).

We should tighten this in 4.0 on the input and output sides. We'll probably keep things a little relaxed on the input side where warranted for ease of use if there is no risk of confusion when it's used.

Issue #368, opened 2020-07-01
Add PEP 484 -- Type Hints


pmhahn:
Python supports type hints to improve static analysis:
<https://www.python.org/dev/peps/pep-0484/>
This helps to identify cases, were wrong types are passed into functions et al by using type checkers like<http://mypy-lang.org/> or <https://github.com/google/pytype/>.

`python-ldap` does not yet seem to contain those annotations nor am I aware of any project doing that.
Those annotation can either we added directly in the library like

> def func(arg: str, val: int = 0) -> Tuple[int, str]

since Python 3.5 or in a Python-2 backward compatible way as

> def func(arg, val=0):  # type: (str, int) -> Tuple[int, str]

As an alternative the annotations can be put into so-called "stub files" names "*.pyi", which are collected by <https://github.com/python/typeshed>, but inline annotations are preferred as they are easier to keep in sync.

I started working on that, but other projects have preempted that
effort. My current work is at
<https://github.com/univention/typeshed/commit/8fc5f796d4ef009e5cf8282575ae304994170b25>,
but it not much more then generating the stubs using `stubgen`.
If someone wants to join that effort that is much appreciated.

This was proposed to the [Mailing List](https://mail.python.org/pipermail/python-ldap/2020q2/004098.html).
@tiran proposed to start the work only after the Python 2 legacy code is removed:
> we are planning to drop support for Python 3.5 and earlier in the next
> version of python-ldap. This would make it simpler to add signatures for
> type hints directly to the code base. Patches are welcome, but please
> wait until we have finalized Python 2 cleanup in master.

encukou:
I would prefer not using `.pyi`. Inline type hints are much more useful, especially for documentation.
Unfortunately, `stubgen` doesn't seem too successful: neither annotating things with `Any`, not specifying `None` return values, nor determinig that constants are `str` or `int`, is very useful.

IMO, type hints are useful for documentation purposes.
I would love to use Sphinx's `autofunction` to generate function signatures in the reference docs; maybe that can be done along with the type annotations?

pmhahn:
> I would prefer not using `.pyi`. Inline type hints are much more useful, especially for documentation.

See my quote from @tiran - type hints will be added directly into the Python code `.py` files *after* the old Python 2 code has been purged.
For now I'm using the `.pyi` files as my project is still Python 2.

[...]

PR #651, opened 2026-09-15
fix(ldap.extop): correct RefreshRequest entryName type

spaceone:
Change the type of `RefreshRequest.entryName` from `str` to `bytes`.

`entryName` is passed to `pyasn1_modules.rfc2251.LDAPDN`, which is based on `pyasn1.type.univ.OctetString` and therefore represents the value as a byte string.

Passing a Unicode string to `OctetString` causes pyasn1 to encode it as ISO-8859-1. LDAP distinguished names use UTF-8 encoding.  Requiring `bytes` avoids implicitly encoding LDAP DNs with the wrong character encoding and makes the expected representation explicit.

Issue: #368
Issue: #647
Fixes: a4110a35


mistotebe:
Should we allow both and encode if `str`? if not, I'm ok with it as is

spaceone:
Supporting both is okay for me. It would also actually fix certain bugs.
In a programmer friendly world data read from "databases" should be decoded before giving them further along, and decoded when putting it back into a protocol element.

spaceone:
support for `str | bytes` has been added by explicitly encode strings as UTF-8.

mistotebe (approved):

PR #652, opened 2026-09-15
refactor: declare TypeAliases explicitly

spaceone:
as it is required by python-typesched linter checks


mistotebe:
OK if we target this for [4.0](https://github.com/python-ldap/python-ldap/milestone/5), since TypeAlias is not in Python 3.9?

spaceone:
if you are okay with it I would make a conditional import e.g. into `_types.py`: ~`compat.py` (or on every use, since `compat` seems unused)~:
```
try:
    from typing import TypeAlias
except ImportError:
     from typing_extensions import TypeAlias
```

spaceone:
@mistotebe I found a fallback strategy for Py3.9 - please have a look at the latest diff. All tests passing.

mistotebe (approved):

PR #653, opened 2026-09-15
Draft: refactor: prefix private TypeVars with "_"

spaceone:
to satisfy flake8-pyi `Y001 Name of private TypeVar must start with _`.

but grr. then one needs to use `stubgen --include-private` which will cause:
```
 stubs/python-ldap/ldap/cidict.pyi:11:5: Y035 "__slots__" in a stub file must have a value, as it has the same semantics as "__slots__" at runtime.
  stubs/python-ldap/ldap/schema/models.pyi:27:5: Y029 Defining __repr__ or __str__ in a stub is almost always redundant
```

Issue: #368


mistotebe:
It has draft in the title but github isn't showing it as draft state. Just checking which it is for review purposes (i.e. have you resolved the stubgen issues or is it pending other stuff?)

spaceone:
jeah, it's a draft because the pre-commit linters of https://github.com/python/typeshed/pull/16395 fight against each other. I would prefer that stubgen works OOTB without including privates. But that's a long path.

I can't set the official draft state here in github somehow. Please don't merge it.

PR #654, opened 2026-09-16
fix(ldap.sasl): allow None in SASL callback values type annotation

spaceone:
it's allowed to pass values e.g. `authzid`/`authcid` as None, which will default to the SASL default value or an empty string:

```python
cb_result: Union[str, bytes] | None = self.cb_value_dict.get(cb_id)
if cb_result is None:
    cb_result = defresult or ''
```

Issue #368


spaceone:
Feel free to reject. I will adjust all callers then, that only empy strings are passed instead.

mistotebe:
You'd have to change it to Mapping to let the concrete implementations pass typecheck, but including None in the passed dict sounds strange so I think I'd prefer you passed `''` as empty or better just not passed those in at all? It's not a -1 from me, more like a question mark. I'll let @droideck weigh in as well

PR #655, opened 2026-09-16
chore: replace ignore[no-untyped-def] anotations with Any

spaceone:
otherwise mypy's `stubgen´ will drop the comments causing:

> Function is missing a type annotation for one or more parameters  [no-untyped-def]


mistotebe (approved):
LGTM!

PR #656, opened 2026-09-22
[3.4 backport] feat(ldap.dn): Add support for different formats in `ldap.dn.dn2str()` via flags

spaceone:
Backport of https://github.com/python-ldap/python-ldap/pull/466/ (https://github.com/python-ldap/python-ldap/issues/257) and https://github.com/python-ldap/python-ldap/pull/634.

I need to maintain these patches anyway in a branch, therefor I can also add this as Pull Request and ask again to include it:
I don't see why it shouldn't be part of a 3.4 release - it is 100% backwards compatible, faster, more secure to use.
And when is 3.5 going to be released? This year? And with 3.5 and major API adjustments it is not suitable for certain Debian releases anymore. Meaning: A feature which was request in 2019 will be available in 2029.


mistotebe:
On the topic of 3.5, I secretly aimed for last week but trying to avoid swamping Simon and yourself with new code and other work trying to get in the way meant that has slipped. Now that the big things are in and once everyone agrees it's in good shape (thanks for all the typing cleanups BTW!), we can get it out.

On trying to get it into 3.4, not sure yet. But also hopefully not urgent to decide? 3.5 is blocking some stuff LDAPCon-wise so I'll prioritise that :) (whereas rushing things into a stable release feels like a Bad Idea(TM)!)

